PR #5, opened 2018-04-30
Fix Gopax trade equals method

grandizzy:


reverendus:
I think the `__hash__` method (aka `hashCode()` ;)) still references the nonexistent field.
Could you please remove it as well?

PR #10, opened 2018-05-11
DDEX: fix is_sell in get_trades()

bargst:


pyexchange/ddex.py
```diff
@@ -226,7 +228,7 @@ def get_trades(self, pair: str, page_number: int = 1) -> List[Trade]:
         trades = list(map(lambda item: Trade(trade_id=item['transactionId'],
                                              timestamp=int(item['executedAt']/1000),
                                              pair=pair,
-                                             is_sell=False if item['taker'] == item['buyer'] else True,
+                                             is_sell= item['buyer'] != tradingAddress,
```
reverendus: Aren't you afraid they will stop sending addresses lowercase one day...? You could use the `Address` class, doing `Address(item['buyer') != Address(self.web3.eth.defaultAccount)` will always work regardless of case.

bargst: I just update the PR with your proposed change, thks !

PR #12, opened 2018-06-01
HitBTC: initial support

bargst:


pyexchange/hitbtc.py
```diff
@@ -0,0 +1,299 @@
...
+    def sell_to_buy_price(self) -> Wad:
+        return self.price
+
+    @property
+    def buy_to_sell_price(self) -> Wad:
+        return self.price
+
+    @property
+    def remaining_buy_amount(self) -> Wad:
+        return self.amount*self.price if self.is_sell else self.amount
```
reverendus: What is `self.amount`? Is it the original amount or the remaining amount?

pyexchange/hitbtc.py
```diff
@@ -0,0 +1,299 @@
...
+    def _auth_post(self, resource: str, params: dict):
+        return self._http(requests.post, resource, params, auth=True)
+
+    def _auth_delete(self, resource: str):
+        return self._http(requests.delete, resource, auth=True)
+
+    def _date_to_timestamp(self, date: str) -> float:
+        # '2018-06-01T14:20:50.497Z'
+        dt = datetime.strptime(date, '%Y-%m-%dT%H:%M:%S.%fZ')
+        return dt.replace(tzinfo=timezone.utc).timestamp()
```
reverendus: No idea why you have to do that... But I don't mind as long as it works :)

pyexchange/hitbtc.py
reverendus: As the logic expects it to be the remaining amount.

PR #15, opened 2018-06-22
 Add page number to get_trades() and get_all_trades()  

bargst:


pyexchange/bibox.py
```diff
@@ -338,7 +338,7 @@ def get_trades(self, pair: str, page_number: int = 1, retry: bool = False) -> Li
 
         return sort_trades(trades)
 
-    def get_all_trades(self, pair: str, retry: bool = False) -> List[Trade]:
+    def get_all_trades(self, pair: str, page_number: int = 1, retry: bool = False) -> List[Trade]:
```
reverendus: Isn't it a bit misleading that you expose a parameter which doesn't really impact the way this function behaves? Maybe we should add `assert(page_number == 1)` to all exchange implementations for which we do not support paging? This way we at least offer protection against such mistakes.

Another thing is that I always tried to have an `assert(isinstance(...))` for all function parameters. Now it's missing for the new `page_number` parameter. I would prefer this to stay consistent i.e. either we assert on the type of all parameters or we completely scrape these assertions. My suggestion would be to keep them as they've already saved my ass once or twice...

PR #16, opened 2018-09-20
Add Ercdex (0xV2 version) to pyexchange

bargst:


bargst:
I added get_balances() to the API

PR #18, opened 2018-12-14
Add Kucoin support

grandizzy:


pyexchange/kucoin.py
```diff
@@ -0,0 +1,269 @@
...
+                     is_sell=trade['direction'] == 'SELL',
+                     price=Wad.from_number(trade['dealPrice']),
+                     amount=Wad.from_number(trade['amount']))
+
+    @staticmethod
+    def from_list(pair, trade):
+        # [1544564526000, 'SELL', 25.0005, 0.0614088, 1.5352507, '5c102f2d335e7e08134edd77']
+        return Trade(trade_id=None,
+                     order_id=trade[5],
+                     timestamp=float(trade[0]),
```
reverendus: Didn't we use `int`s for timestamps for other exchanges..?

pyexchange/kucoin.py
```diff
@@ -0,0 +1,269 @@
...
+        assert(isinstance(is_sell, bool))
+        assert(isinstance(price, Wad))
+        assert(isinstance(amount, Wad))
+
+        side = self.client.SIDE_SELL if is_sell else self.client.SIDE_BUY
+
+        coins = pair.split("-")
+
+        price = self._get_precision(coins[1]) % float(price)
+        amount = self._get_precision(coins[0]) % float(price)
```
reverendus: are you sure it shouldn't say `float(amount)` here??

pyexchange/kucoin.py
```diff
@@ -0,0 +1,269 @@
...
+        assert(isinstance(pair, str))
+
+        result = self.client.get_recent_orders(pair)
+        print(result)
+
+        return list(map(lambda item: Trade.from_list(pair, item), result))
+
+    @staticmethod
+    def _get_precision(coin):
+        return {
```
reverendus: :/ is there no way to fetch it from the API dynamically..?

pyexchange/kucoin.py
grandizzy: API rejects if they do not follow an exact precision, I determined precision by testing different pairs as didn't find any documentation for

pyexchange/kucoin.py
grandizzy: didn't find such :(

pyexchange/kucoin.py
grandizzy: change in 2nd commit

requirements.txt
```diff
@@ -3,3 +3,4 @@ web3 == 4.5.0
 requests == 2.18.4
 python-dateutil == 2.6.1
 websockets == 6.0.0
+python-kucoin == 0.1.12
```
bargst: I don't really like the fact to add a specific exchange dependency here

bargst: Is there a way to make it optional ?

grandizzy: per our discussion in chat, do you think that's OK to go with this approach for the moment? thanks!

PR #20, opened 2018-12-19
Add Coinbase support

grandizzy:


pyexchange/coinbase.py
```diff
@@ -0,0 +1,295 @@
...
+
+    @staticmethod
+    def _get_precision(coin):
+        return {
+            'ETH': "%.7f",
+            'USDC': "%.2f"
+        }[coin]
+
+
+def http_response_summary(response) -> str:
```
reverendus: Why have you copied that method from `pymaker`...?

grandizzy: oops, fixed with second commit

PR #83, opened 2019-07-19
Uniswap add eth_to_token_swap_input and token_to_eth_swap_input

devenoges:
Implemented basic swaps, will add min_eth and min_token args.

What do you think about making deadline value configurable?

<insert-animal-pic-here/>


KentonPrescott:
@EdNoepel once this PR is merged, I can make one that includes functionality to convert tokens -> tokens

PR #96, opened 2019-09-25
Add Kraken support

grandizzy:


pyexchange/kraken.py
```diff
@@ -0,0 +1,301 @@
...
+        assert(isinstance(pair, str))
+        assert(isinstance(page_number, int))
+        assert(page_number == 1)
+
+        result = self._http_authenticated("POST", f"/0/private/TradesHistory", {})
+
+        trades = []
+
+        for trade_id, trade in result['trades'].items():
+            if trade['pair'] == pair and trade['ordertype'] == 'limit':
```
ith-harvey: No need to specify `trade['ordertype'] == 'limit':` all orders should already be limit orders no?

pyexchange/kraken.py
```diff
@@ -0,0 +1,301 @@
...
+
+        headers = {
+            'API-Key': self.api_key,
+            'API-Sign': sigdigest.decode()
+        }
+
+        return self._result(requests.request(method=method,
+                                             url=f"{self.api_server}{resource}",
+                                             data=body,
+                                             headers=headers,
```
ith-harvey: move headers block into here just like in CB ->
```
  return self._result(requests.request(method=method,
                                             url=f"{self.api_server}{resource}",
                                             data=data,
                                             headers={
                                                 'Content-Type': 'Application/JSON',
                                                 'CB-ACCESS-SIGN': signature_b64,
                                                 'CB-ACCESS-TIMESTAMP': timestamp,
                                                 'CB-ACCESS-KEY': self.api_key,
                                                 'CB-ACCESS-PASSPHRASE': self.password
                                             },
                                             timeout=self.timeout))
```

pyexchange/kraken.py
grandizzy: there could be multiple types of orders

market
    limit (price = limit price)
    stop-loss (price = stop loss price)
    take-profit (price = take profit price)
    stop-loss-profit (price = stop loss price, price2 = take profit price)
    stop-loss-profit-limit (price = stop loss price, price2 = take profit price)
    stop-loss-limit (price = stop loss trigger price, price2 = triggered limit price)
    take-profit-limit (price = take profit trigger price, price2 = triggered limit price)
    trailing-stop (price = trailing stop offset)
    trailing-stop-limit (price = trailing stop offset, price2 = triggered limit offset)
    stop-loss-and-limit (price = stop loss price, price2 = limit price)
    settle-position

ith-harvey: yes but what will we be placing... Only `limit` orders right

grandizzy: Correct, so we just want to fetch limit orders

pyexchange/kraken.py
grandizzy: done

PR #104, opened 2019-12-06
Akash leverj pyexchange changes20191206

mitakash:


EdNoepel:
`pyexchange` has some serious unit test deficiencies.  I assume the cost of implementing tests (which properly recreate relevant exchange contracts and sidechain) outweighs the benefits of improving code coverage.

pyexchange/leverj.py
```diff
@@ -375,39 +376,50 @@ def approve_token(self, token_address: str, amount: int) -> Transact:
...
-        assert(isinstance(gluon_block_number, None) or isinstance(gluon_block_number, int))
 
         custodian_account = self.address
         app_id = leverjobj.get_spot_exchange_id()
         if gluon_block_number is None:
-            gluon_block_number = leverjobj._http_authenticated("GET", "/api/v1", f"/plasma/{app_id}", None)['number'] +2
-            Transact(self, self.web3, self.abi, self.address, self._contract, "depositEther",[app_id], {'value': int(amount.value)}).transact()
-            return gluon_block_number
+            gluon_block_number = leverjobj._http_authenticated("GET", "/api/v1", f"/plasma/{app_id}", None)['number'] + 2
+            receipt = Transact(self, self.web3, self.abi, self.address, self._contract, "depositEther",[app_id], {'value': int(amount.value)}).transact()
```
grandizzy: do we want just to return receipt or to check here if transaction completed ok / other details?

pyexchange/leverj.py
```diff
@@ -375,39 +376,50 @@ def approve_token(self, token_address: str, amount: int) -> Transact:
...
-        assert(isinstance(gluon_block_number, None) or isinstance(gluon_block_number, int))
 
         custodian_account = self.address
         app_id = leverjobj.get_spot_exchange_id()
         if gluon_block_number is None:
-            gluon_block_number = leverjobj._http_authenticated("GET", "/api/v1", f"/plasma/{app_id}", None)['number'] +2
-            Transact(self, self.web3, self.abi, self.address, self._contract, "depositToken",[app_id, token_address, amount], {}).transact()
-            return gluon_block_number
+            gluon_block_number = leverjobj._http_authenticated("GET", "/api/v1", f"/plasma/{app_id}", None)['number'] + 2
+            receipt = Transact(self, self.web3, self.abi, self.address, self._contract, "depositToken",[app_id, token_address, amount], {}).transact()
```
grandizzy: same as above, should we just return receipt?

mitakash: We need to return both, gluon_block_number tells us where we are in the process, the hash in the receipt is eventually sent back into post_pending_tx_hash to let let leverj know we just sent something and to look for that transaction hash in pending transactions on ethereum block chain.

pyexchange/leverj.py
mitakash: we need to return receipt in it's full form to be checked by inventory service with gluon block to know where we are in the processing

PR #105, opened 2019-12-14
Akash middle account change20191214

mitakash:


pyexchange/leverj.py
```diff
@@ -442,24 +456,25 @@ def claim_funds(self, leverjobj: LeverjAPI, asset: str, quantity: int, gluon_blo
...
-            custodian_account = self.address
-            self.logger.info(f"ethereum_account: {ethereum_account}, custodian_account: {custodian_account}, asset: {asset}")
-            response = leverjobj._http_authenticated("GET", "/api/v1", f"/plasma/{app_id}/evmparams/withdrawals/account/{ethereum_account}/asset/{asset}", None)
-            response_app_id = int(response[0])
-            response_bytes = response[1]
             current_block = leverjobj._http_authenticated("GET", "/api/v1", f"/plasma/{app_id}", None)['number']
-
-            if (leverjobj._http_authenticated("GET", "/api/v1", f"/plasma/{app_id}", None)['number'] >= gluon_block_number):
+            if current_block >= gluon_block_number:
+                #leverjobj.web3.eth.defaultAccount = leverjobj.account_id
```
ith-harvey: If not being used, best to remove this line.

pyexchange/leverj.py
```diff
@@ -442,24 +456,25 @@ def claim_funds(self, leverjobj: LeverjAPI, asset: str, quantity: int, gluon_blo
...
+                response_app_id = int(response[0])
+                response_bytes = response[1]
                 self.logger.info(f"finally gluon_block_number reached {gluon_block_number} and we are running final transact")
-                Transact(self, self.web3, self.abi, self.address, self._contract, "withdraw",[response_app_id, response_bytes], {}).transact()
+                Transact(self, self.web3, self.abi, self.address, self._contract, "withdraw",[response_app_id, response_bytes], {}).transact(from_address=self.middle_account)
                 return None
-        
-        self.logger.info(f'does not look like gluon_block_number reached {gluon_block_number} and we are currently at {leverjobj._http_authenticated("GET", "/api/v1", f"/plasma/{app_id}", None)["number"]}')
+
+        current_block = leverjobj._http_authenticated("GET", "/api/v1", f"/plasma/{app_id}", None)['number']
```
ith-harvey: noticing you do this earlier inside of the else statement above ->
`current_block = leverjobj._http_authenticated("GET", "/api/v1", f"/plasma/{app_id}", None)['number']`
Is it possible to put this call in the global scope? that way you can call it once and then use the variable again later instead of calling it twice?

pyexchange/leverj.py
mitakash: removed

pyexchange/leverj.py
mitakash: above has been moved to one line now for current block calculation

PR #108, opened 2019-12-19
Add e toro

MikeHathaway:


MikeHathaway:
Also added a separate test file, but must have accidentally carried over the regex for updating the names to that file. I'll revert the change to the okex tests

ith-harvey:
cool, also worth waiting on having the team review untill we have an account we can use to test placing order etc... that way people can review all at once and don't have to double back. I'll double back with Joe today to check on acc status

MikeHathaway:
Yeah makes sense. We've requested API access but it's still pending approval from the exchange

tests/test_etoro.py
```diff
@@ -0,0 +1,224 @@
...
+        if re.search(r"api\/v1\/instruments", url):
+            return MockedResponse(text=EToroMockServer.responses["markets"])
+        elif re.search(r"api\/v1\/balances", url):
+            return MockedResponse(text=EToroMockServer.responses["balances"])
+        elif re.search(r"api\/v1\/orders", url):
+            return MockedResponse(text=EToroMockServer.responses["orders"])
+        elif re.search(r"\/api\/v1\/trades", url):
+            return MockedResponse(text=EToroMockServer.responses["trades"])
+        else:
+            raise Exception("Unable to match HTTP GET request to canned response")
```
EdNoepel: Although you'll find it all over our code, I recommend not throwing the base `Exception` class.  Makes it harder for consumers to catch anticipated exceptions and avoid catching ones which shouldn't be caught.

tests/manual_test_etoro.py
```diff
@@ -0,0 +1,47 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import sys
+
+from pyexchange.etoro import EToroApi
+from pymaker.numeric import Wad
+
+etoro = EToroApi('https://7vriaeqd.hft.etorox.com', sys.argv[1], open('./.etoro-unencrypted-key', 'r').read(), 9.5)
+# etoro = EToroApi('https://7vriaeqd.hft.etorox.com', sys.argv[1], open(sys.argv[2], "r",encoding='utf-8').read(), 9.5)
```
EdNoepel: Parameterizing the URL and auth credentials would make it easier for others to test.

pyexchange/etoro.py
```diff
@@ -0,0 +1,332 @@
...
+
+import logging
+from pprint import pformat
+from typing import Optional, List
+
+import dateutil.parser
+import requests
+import time
+import hmac
+import hashlib
```
EdNoepel: Where are `hmac` and `hashlib` used?

tests/test_etoro.py
```diff
@@ -0,0 +1,224 @@
...
+            instrument_id="ethusdc",
+            is_sell=False,
+            price=price,
+            amount=amount,
+            remaining_amount=remaining_amount
+        )
+        assert(order.price == order.sell_to_buy_price)
+        assert(order.price == order.buy_to_sell_price)
+        # assert(order.remaining_buy_amount == amount-remaining_amount)
+        # assert(order.remaining_sell_amount == (amount-remaining_amount)*price)
```
EdNoepel: Perhaps you could replace these comments with `assert(order.remaining_amount == remaining_amount)`.

pyexchange/etoro.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def __repr__(self):
+        return pformat(vars(self))
+
+
+class EToroApi(PyexAPI):
+    """eToro API interface.
+
+    Developed according to the following manual:
+    <https://etorox.github.io/dxocs/#/>.
```
EdNoepel: This 404's on me.  Could you confirm?

pyexchange/etoro.py
```diff
@@ -0,0 +1,332 @@
...
+
+    @property
+    def remaining_sell_amount(self) -> Wad:
+        return self.remaining_amount if self.is_sell else self.remaining_amount*self.price
+
+    def __repr__(self):
+        return pformat(vars(self))
+
+    @staticmethod
+    def to_order(item):
```
EdNoepel: Since this factory method sits in the `Order` class, I'd expect a name like `from_message`.  Current name suggests we're trying to convert an `Order` to another object.

pyexchange/etoro.py
```diff
@@ -0,0 +1,332 @@
...
+        return Order(order_id=item['id'],
+                     timestamp=int(time.time()), # No timestamp or created at information is returned as part of get_orders()
```
EdNoepel: This should be UTC, not local time.  We may even want to remove the field altogether if their API doesn't provide it.

pyexchange/etoro.py
```diff
@@ -0,0 +1,332 @@
...
+    def remaining_buy_amount(self) -> Wad:
+        return self.remaining_amount*self.price if self.is_sell else self.remaining_amount
+
+    @property
+    def remaining_sell_amount(self) -> Wad:
+        return self.remaining_amount if self.is_sell else self.remaining_amount*self.price
+
+    def __repr__(self):
+        return pformat(vars(self))
+
```
EdNoepel: No `__hash__` method?

MikeHathaway: Added __hash__ method

pyexchange/etoro.py
MikeHathaway: Makes sense. Updated name to from_message

pyexchange/etoro.py
MikeHathaway: The place_order response object from their API doesn't include a timestamp of any sort. They do expose a public API for getting their server time which I could use to replace this.

tests/test_etoro.py
MikeHathaway: Replaced with your recommendation.

pyexchange/etoro.py
MikeHathaway: They weren't. Removed in latest commit

pyexchange/etoro.py
MikeHathaway: Had a typo - fixed

pyexchange/etoro.py
```diff
@@ -0,0 +1,302 @@
...
+        params['correlationId'] = str(uuid.uuid4())
+
+        nonce = str(uuid.uuid4())
+        timestamp = str(int(time.time() * 1000))
+
+        print(self.secret_key)
+        signature = base64.b64encode(hmac.new(self.secret_key.encode('utf-8'), f"{nonce}{timestamp}".encode('utf-8'), hashlib.sha256).digest())
+
+        headers = {
+            "user-agent": "mm@liquidityproviders.io",
```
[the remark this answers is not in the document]
MikeHathaway: Added "account" to the constructor

tests/test_okex.py
```diff
@@ -26,7 +26,7 @@
 from pyexchange.okex import Trade
 from pyexchange.okex import OKEXApi
 
-# Models HTTP response, produced by OkexMockServer
+# Models HTTP response, produced by EToroMockServer
```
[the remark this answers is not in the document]
MikeHathaway: Accidentally carried over Regex. Fixed it

tests/test_etoro.py
MikeHathaway: Updated to ValueError and added the test URL and Data field as part of the message

tests/manual_test_etoro.py
MikeHathaway: Reverted back to `sys.argv` for host and credentials.

ith-harvey:
@MikeHathaway github is bugging out so leaving this here... inside the Etoro class can you leave a note here specifying that the software requires an unencrypted `secret_key` and to do so the secret key Etoro provides needs to be unencrypted with an empty " "?

EdNoepel:
> @MikeHathaway github is bugging out so leaving this here... inside the Etoro class can you leave a note here specifying that the software requires an unencrypted `secret_key` and to do so the secret key Etoro provides needs to be unencrypted with an empty " "?

He did include a comment about that here: https://github.com/makerdao/pyexchange/pull/108/files#diff-1f24dbd9e0103e7cf2eafdcb5065862aR165

PR #113, opened 2020-01-21
Update Get Orders to indicate state is a required param

MikeHathaway:


MikeHathaway:
> Please update the `test_get_orders` unit test accordingly.

Just pushed up a fix that tests with a default param of "open" for all orders of that pair.

PR #116, opened 2020-01-27
Add bitso

MikeHathaway:


pyexchange/bitso.py
```diff
@@ -0,0 +1,339 @@
...
+            return self._result(requests.request(method=method,
+                                             url=url,
+                                             headers=headers,
+                                             json=data,
+                                             timeout=self.timeout))
+    @staticmethod
+    def _result(result) -> dict:
+
+        if not result.ok:
+            raise Exception(f"Bitso API invalid HTTP response: {http_response_summary(result)}")
```
EdNoepel: Raising the base `Exception` class is frowned upon because it makes it more difficult for calling code to catch only expected exceptions.  I know the other exchanges in the library do this, but maybe we should break consistency rather than accumulate technical debt.

MikeHathaway: Replaced with ValueError, as discussed.

PR #118, opened 2020-01-30
Add korbit

MikeHathaway:


pyexchange/korbit.py
```diff
@@ -0,0 +1,374 @@
...
+            "id": order_id
+        }
+
+        result = self._http_authenticated_request("POST", f"/v1/user/orders/cancel", data)
+        if result[0]["status"] == "success":
+            return True
+        else:
+            return False
+
+    # TODO: finish implementation
```
EdNoepel: Is this a stale comment, or is something actually incomplete?

MikeHathaway: Stale comment, I will remove it

tests/test_korbit.py
```diff
@@ -0,0 +1,205 @@
...
+        for order in response:
+            assert(isinstance(order.is_sell, bool))
+            assert(Wad(order.price) > Wad(0))
+        TestKorbit.check_orders(response)
+
+    def test_order_placement_and_cancellation(self, mocker):
+        pair = "dai_krw"
+        side = "ask"
+        mocker.patch("requests.request", side_effect=KorbitMockServer.handle_request)
+        order_id = self.korbit.place_order(pair, True, 1500, Wad.from_number(10))
```
EdNoepel: Innit that a `Wad` now?

pyexchange/korbit.py
```diff
@@ -0,0 +1,361 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+from pprint import pformat
+from pyexchange.api import PyexAPI
+import hmac
```
EdNoepel: `hmac`, `hashlib`, `base64', and `dateutil.parser` seem to be unused imports.

MikeHathaway: Removed unused imports

pyexchange/korbit.py
```diff
@@ -361,4 +364,19 @@ def _format_pair_string(self, pair: str) -> str:
         if '-' in pair:
             return "_".join(pair.split('-')).lower()
         else:
-            return pair.lower()
\ No newline at end of file
+            return pair.lower()
+
+    def _choose_nonce(self) -> int:
```
EdNoepel: Please explain why threadsync is necessary.  Each method which requires a nonce is making a synchronous call, so concurrency is unexpected.

pyexchange/korbit.py
```diff
@@ -361,4 +364,19 @@ def _format_pair_string(self, pair: str) -> str:
...
+        with self.last_nonce_lock:
+            timed_nonce = int(time.time()*1000)
+            time.sleep(0.1)
+
+            if self.last_nonce + 1 > timed_nonce:
+                self.logger.info(f"Wanted to use nonce '{timed_nonce}', but last nonce is '{self.last_nonce}'")
```
EdNoepel: Please make these a single DEBUG-level log message.

MikeHathaway: Done

pyexchange/korbit.py
MikeHathaway: This was added to resolve too many requests being fired off by the keeper at the same time, which have conflicting nonces from the perspective of the Korbit API, and then are subsequently rejected.

The approach is similar to that used in Kraken and Liquid: https://github.com/makerdao/pyexchange/commit/affec1616cc69edd08b63225b5fb9c1b88a6a423

PR #120, opened 2020-02-20
Add okcoin

MikeHathaway:


pyexchange/okcoin.py
```diff
@@ -0,0 +1,67 @@
...
+        assert(isinstance(password, str))
+
+        super().__init__(api_server, api_key, secret_key, password, timeout)
+
+    def get_markets(self):
+        return self._http_get(f"/api/spot/v3/instruments", "")
+
+    # Retrieve address for specified tokens Funding Account
+    def get_deposit_address(self, currency: str = "eth") -> str:
+        return self._http_get(f"/api/account/v3/deposit/address", f"currency={currency}", requires_auth=True, has_cursor=False)
```
EdNoepel: Does this work?  If so, please adjust comment in manual test.  If not, recommend commenting out.

MikeHathaway: Thanks for catching that. It works and i've cleaned up the manual tests as well

PR #123, opened 2020-03-25
Add dydx

MikeHathaway:


pyexchange/dydx.py
```diff
@@ -0,0 +1,261 @@
...
+        return self.client.get_pairs()['pairs']
+
+    def get_pair(self, pair: str):
+        assert (isinstance(pair, str))
+        return next(filter(lambda symbol: symbol['name'] == pair, self.get_markets()))
+
+    def _balances_to_list(self, balances) -> List:
+        balance_list = []
+
+        for token, balance in enumerate(balances):
```
grandizzy: can we reuse dydx constants as below e.g. ```consts.MARKET_ETH```

pyexchange/dydx.py
```diff
@@ -0,0 +1,261 @@
...
+        assert (isinstance(price, float))
+        assert (isinstance(amount, float))
+
+        side = 'SELL' if is_sell else 'BUY'
+
+        self.logger.info(f"Placing order ({side}, amount {amount} of {pair},"
+                         f" price {price})...")
+
+        ## Need to retrieve the market_id used by a given token as all trades in DyDx use Wei as standard unit.
+        market_id = 0
```
grandizzy: maaybe we should reuse pymaker's `model.Token` class and call normalize amount to get the price and make sure everything is consistent

pyexchange/dydx.py
```diff
@@ -0,0 +1,261 @@
...
+import dateutil.parser
+
+import dydx.constants as consts
+import dydx.util as utils
+
+
+class Order:
+    def __init__(self,
+                 order_id: str,
+                 timestamp: int,
```
ith-harvey: None of the other pyexchange Order classes seem to have the `timestamp` attribute... Is it necessary that we use it?

I guess it cant hurt us to keep it if it's returned from Dydx as long as there aren't any issues when the keeper posts the order to the `order-history-service`

pyexchange/dydx.py
```diff
@@ -0,0 +1,261 @@
...
+    def remaining_sell_amount(self) -> Wad:
+        return self.amount if self.is_sell else self.amount * self.price
+
+    def __repr__(self):
+        return pformat(vars(self))
+
+    @staticmethod
+    def to_order(item: list, pair: str):
+        return Order(order_id=item['id'],
+                     timestamp=int(dateutil.parser.parse(item['createdAt']).timestamp()),
```
ith-harvey: see timestamp comment above

tests/manual_test_ddex.py
```diff
@@ -20,6 +20,7 @@
 import time
 
 from web3 import Web3, HTTPProvider
+import web3.utils
```
ith-harvey: this will end up being `_utils` in the updated web3 version right?

tests/test_dydx.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @staticmethod
+    def handle_get_trades(**kwargs):
+        return MockedResponse(text=DydxMockServer.responses["trades"]).json()
+
+class TestDydx:
+    def setup_method(self):
+        self.dydx = DydxApi(
+            "https://mainnet.infura.io/v3/cde88afeb2d9475aa8a7b3813e7386cf",
+            "dcba44978751342a68e81b0e487de87e52720f6f94792cc237045bce0f9d05fc"
```
ith-harvey: is this a fake private key or the real deal?

pyexchange/airswap.py
```diff
@@ -16,29 +16,19 @@
...
 from web3 import Web3
 from web3._utils.events import get_event_data
 
-from pymaker import Contract, Address, Transact, Receipt
+from pymaker import Contract, Address
 from pymaker.numeric import Wad
-from pymaker.token import ERC20Token
-from pymaker.util import int_to_bytes32, bytes_to_int
+from pymaker.util import bytes_to_int
+from pymaker.util import http_response_summary
```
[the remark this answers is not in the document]
MikeHathaway: those imports are unused

pyexchange/dydx.py
MikeHathaway: Okex, Okcoin, eToro, Bitso, and Korbit all have a `timestamp` attribute, and they've been able to post successfully to `order-history-service`. Given it's returned from the exchange, and harmless this is a form of future proofing in case we ever want to use order data

pyexchange/dydx.py
MikeHathaway: answered above

pyexchange/dydx.py
MikeHathaway: what would be the purpose of this? This code is needed as the other services expect balance to provide as an array, not as a dictionary. I tried, but it wasn't clear how this could be made reusable so I didn't pull it out into a separate function.

tests/manual_test_ddex.py
MikeHathaway: yep, good catch. I didn't push up the manual test file ive been using locally

tests/test_dydx.py
MikeHathaway: real private key, but for an unused account

pyexchange/dydx.py
MikeHathaway: Nvm, I think i understand. I just updated the comparison to use consts instead of an int primitive as - e.g. `if token == consts.MARKET_ETH:`

pyexchange/dydx.py
MikeHathaway: Looked into this, but I don't think that will work. DyDx only accepts a specific `wei` formatting, without any decimals. `Token.normal_amount` doesn't strip out the decimals or convert to `wei`. That being said, it might be worth the effort to add a method to `pymaker.model.Token` such as `convert_to_wei`

pyexchange/airswap.py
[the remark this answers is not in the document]
EdNoepel: I appreciate the cleanup.

pyexchange/dydx.py
EdNoepel: Timestamp will be helpful for PnL.

tests/manual_test_dydx.py
```diff
@@ -0,0 +1,25 @@
+# This file is part of Maker Keeper Framework.
+#
+# Copyright (C) 2019 grandizzy
```
EdNoepel: Correct author?  Seems silly that we have a header in any of these `manual_test`s regardless.

pyexchange/dydx.py
```diff
@@ -62,12 +62,14 @@ def buy_to_sell_price(self) -> Wad:
 
     @property
     def remaining_buy_amount(self) -> Wad:
-        return self.amount * self.price if self.is_sell else self.amount
+        amount = Wad.from_number(from_wei(abs(int(float(self.amount))), 'ether'))
```
EdNoepel: Where is `from_wei` defined?

MikeHathaway: good catch, added the import from eth_utils

PR #127, opened 2020-04-24
Update kucoin-python; backport Order and Trade models to dydx

MikeHathaway:


MikeHathaway:
All Dydx changes were retested manually as well

PR #128, opened 2020-04-27
Update DyDx place order code to account for updated precision requirements

MikeHathaway:


pyexchange/dydx.py
```diff
@@ -151,7 +151,7 @@ def place_order(self, pair: str, is_sell: bool, price: float, amount: float) ->
         created_order = self.client.place_order(
             market=pair,  # structured as <MAJOR>-<Minor>
             side=side,
-            price=Decimal(price),
+            price=round(Decimal(price), 2),
```
grandizzy, line 154: is this something we can retrieve from DyDx API / python client?

MikeHathaway: Yep, code has been added to retrieve that and `market_id` from the new `get_markets()` endpoint

PR #130, opened 2020-04-30
Bug fixes for USDC Decimal handling in DyDx

MikeHathaway:


pyexchange/dydx.py
```diff
@@ -140,13 +166,20 @@ def place_order(self, pair: str, is_sell: bool, price: float, amount: float) ->
...
         market_info = self.get_pair(pair)
         tick_size = abs(Decimal(market_info['minimumTickSize']).as_tuple().exponent)
-        # As market_id is only used for amount, use baseCurrency instead of quoteCurrency
+        # As market_id is used for amount, use baseCurrency instead of quoteCurrency
         market_id = market_info['baseCurrency']['soloMarketId']
+        quote_market_id = market_info['quoteCurrency']['soloMarketId']
+
+        # if USDC adjust price for decimals difference
+        if quote_market_id == consts.MARKET_USDC:
+            price = Decimal(f"{price}e-12")
```
EdNoepel, line 175: `market_info['baseCurrency']['decimals']` will provide the decimal precision (see https://docs.dydx.exchange/#/api?id=get-v2marketsmarket).  Since you're already making the call to get tick size, can't you use that data here and avoid a conditional on token name?

MikeHathaway: Yes, that would be cleaner. I was re-purposing old code, but will adjust accordingly.

MikeHathaway: I think a conditional of some sort would still be necessary - `market_info['baseCurrency']['decimals']` returns the decimal precision, but the price needs to be expressed in terms of 18 decimal wei, so it would have be multiplied by `10 ** -12`

MikeHathaway: Yes that would be cleaner. I've adjusted it accordingly

MikeHathaway:
> I'd prefer to let dydx market info to determine number of decimal places, and avoid switching on token name wherever possible. Maybe `market_info` could be passed into the `DydxOrder` and `DydxTrade` factory methods?
> 
> A cache would avoid the need to repeatedly query the same market info.

I added a `self.market_info` attribute to the class created on instantiated, but it would require pair information as `get_markets()` returns a dict keyed by pair - which means decimals usage in `get_balances` couldn't easily be replaced without breaking the PyEx interface.

If there's a better approach to this I would be open to suggestions

PR #131, opened 2020-05-04
handle dydx pending balance change edge case

MikeHathaway:
This change decreases the available balance if an order is in the process of being filled, thereby resolving an edge case that led to use of margin.


mitakash:
Looks good to me!

PR #132, opened 2020-05-08
Add erisx

MikeHathaway:
`get_all_trades` is still in progress, but otherwise, main functionality is in place for spot integration


pyexchange/fix.py
```diff
@@ -0,0 +1,292 @@
...
+        self.session_loop = None
+        self.last_msg_sent = None
+        self.logging_out = False
+
+        self.write_queue = queue.Queue()
+        self.application_messages = queue.Queue()
+
+    async def _read_message(self):
+        """Reads the next message from the server"""
+        # await self.lock.acquire()
```
EdNoepel, line 87: I need to spend some time with the `asyncio` docs confirming these locks aren't really needed.

pyexchange/fix.py
```diff
@@ -0,0 +1,292 @@
...
+        try:
+            message = None
+            # logging.debug("reading")
```
EdNoepel, line 90: I should push a commit removing such debugging statements.

tests/mock_fix_server.py
```diff
@@ -0,0 +1,91 @@
...
+
+import asyncio
+import os
+import simplefix
+import threading
+
+from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
+
+
+class MockFixServer:
```
EdNoepel, line 26: This implementation should be thrown away.  I considered creating a new one which opens the server side socket and waits, but it'd be expensive.  Buying a solution like _VeriFIX_ is also expensive.  Less discuss again in standup.

pyexchange/erisx.py
```diff
@@ -0,0 +1,400 @@
...
+    ErisX documentation available here: https://www.erisx.com/wp-content/uploads/2020/03/ErisX-FIX-4.4-Spec-V3.3.pdf
+    """
+
+
+    logger = logging.getLogger()
+    timeout = 5
+
+    def __init__(self, fix_trading_endpoint: str, fix_trading_user: str,
+                 fix_marketdata_endpoint: str, fix_marketdata_user: str, password: str,
+                 clearing_url: str, api_key: str, api_secret: str, web_api_only: bool):
```
EdNoepel, line 75: Should `web_api_only` be implemented by letting the user pass `clearing_url=None`?  Note this seems to have broken `manual_test_erisx.py`.

MikeHathaway: ah, i only updated `manual_test_erisx.py` locally. Thanks for the suggestion, that would be much cleaner. In the same vein, could perhaps use `""` instead of `None` so we could preserve the `isinstance` checks? And then check against that and only pass non empty string if needed

EdNoepel: Although they're falsey, I'm not a fan of empty strings.  We usually do `assert isinstance(endpoint, str) or endpoint is None`.

pyexchange/fix.py
EdNoepel: `asyncio.StreamReader` and `StreamWriter` are not threadsafe.  Reads always initiate from the session thread, so no problem there.  Writes _usually_ initiate from a (threadsafe) standard library [Queue](https://docs.python.org/3.6/library/queue.html?highlight=queue#module-queue).  The only exception is `logout` which blocks on the caller's event loop while `_session_proc` keeps churning the `write_queue`.  As such, I removed the (commented-out) one from `_read_message` and uncommented the one in `_write_message`.

pyexchange/erisx.py
```diff
@@ -0,0 +1,400 @@
...
+            self.fix_marketdata = ErisxFix(fix_marketdata_endpoint, fix_marketdata_user, fix_trading_user, password)
+            self.fix_marketdata.logon()
+            self.fix_marketdata_user = fix_marketdata_user
+
+        self.clearing_url = clearing_url
+        self.api_secret = api_secret
+        self.api_key = api_key
+
+        # store the account id used to retrieve trades and balances
+        self.account_id = self.get_account()
```
EdNoepel, line 102: This breaks the unit tests, because the mocker isn't available in `setup_method`, so it tries to hit the real URL.  I'm working on it.

pyexchange/erisx.py
```diff
@@ -0,0 +1,399 @@
...
+    # filter out the response from get markets
+    def get_pair(self, pair):
+        return self.get_markets()[pair]
+
+    def get_account(self):
+        # Call into the /accounts method of ErisX Clearing WebAPI, which provides a balance of each coin.
+        # They also offer a detailed /balances API, which I don't believe we need at this time.
+        response = self._http_post("accounts", {})
+        if "accounts" in response:
+            return response["accounts"][0]["account_id"]
```
EdNoepel, line 131: I doubt we should place any special importance upon the first account returned by the API.  This impacts the constructor, which calls this method.  Most exchange APIs seem to provide data across all accounts; we could do the same here client-side.

pyexchange/erisx.py
```diff
@@ -0,0 +1,399 @@
...
+    # Trade information is only retrieved on a per session basis through FIX (Page 20 of Spec)
+    def get_trades(self, pair: str, page_number: int = 8) -> List[Trade]:
+        response = self._http_post("trades", {"account_id": self.account_id})
+        return list(map(lambda trade: ErisxTrade.from_message(trade), response["trades"]))
+
+    # used to unsubscribe from the marketdata feed used in get_all_tradesx
+    def unsubscribe_marketdata(self, pair: str, request_id: str) -> bool:
+        assert(isinstance(request_id, str))
+
+        message = self.fix_marketdata.create_message(simplefix.MSGTYPE_MARKET_DATA_REQUEST)
```
EdNoepel, line 224: Although I neglected to do so with `get_markets`, I'd recommend moving message construction to the `ErisxFix` class below.

pyexchange/erisx.py
MikeHathaway: are you imaging that as the client would pass an additional argument to the constructor signifying which account they would like to use?

pyexchange/erisx.py
MikeHathaway: just for these marketdata methods, or for all methods that involve FIX message construction?

EdNoepel: After examining `place_order`, I'm having second thoughts about this.  Lots of arguments would need to be duplicated, complicating code maintenance.

Happy to leave it alone for now.  When we do the next integration, we'll more clearly see opportunities to move some `place_order` and `cancel_order` message construction logic to the base `FixEngine` class, and use `ErisxFix` to override the implementation.

pyexchange/erisx.py
EdNoepel: I hadn't been looking at it that way because It seemed undesirable to create three new connections for each account we want to trade with.  But it would make this fit into `market-maker-keeper` nicely, which I suppose is the goal.  The `account_id` could be passed as an arg to `market-maker-keeper`, which offers it to the `ErisxApi` constructor.

pyexchange/erisx.py
MikeHathaway: Yeah that makes sense to me

pyexchange/erisx.py
MikeHathaway: Just updated accordingly with `account_id` arg

pyexchange/fix.py
MikeHathaway: Makes sense, thanks for the fix

pyexchange/erisx.py
```diff
@@ -72,7 +72,7 @@ class ErisxApi(PyexAPI):
 
     def __init__(self, fix_trading_endpoint: str, fix_trading_user: str,
                  fix_marketdata_endpoint: str, fix_marketdata_user: str, password: str,
-                 clearing_url: str, api_key: str, api_secret: str):
+                 clearing_url: str, api_key: str, api_secret: str, account_id: int = 0):
```
EdNoepel, line 75: That makes sense; usually the consumer wants the first on the list.  This allows them to connect even if they don't know what their account_id is.

PR #134, opened 2020-05-13
add pair regex check for sync trades

MikeHathaway:


pyexchange/model.py
```diff
@@ -112,6 +114,9 @@ def __init__(self,
         assert(isinstance(price, Wad))
         assert(isinstance(amount, Wad))
 
+        # Ensure that pair schema matches expectations from sync-trades
+        assert(re.match('[A-Z0-9]+\-[A-Z0-9]+', pair))
```
EdNoepel, line 118: We've got a few symbols (`cDAI` `USStocks` `cBAT`) with lowercase letters.

MikeHathaway: Updated regex to account for those

PR #136, opened 2020-05-14
Update erisx to place post only orders

MikeHathaway:


pyexchange/erisx.py
```diff
@@ -210,7 +213,10 @@ def cancel_order(self, order_id: str, pair: str, is_sell: bool) -> bool:
         self.fix_trading.write(message)
 
         response = self.fix_trading.wait_for_response('8')
-        return True if response.get(150).decode('utf-8') == '4' else False
+        if response.get(150) is not None:
```
MikeHathaway, line 216: Added this to handle an edge case we shouldn't ever see where the keeper was manually killed before canceling, and then currency pair was switched.

EdNoepel: For each cancel, you'll likely get two responses, the first a 150=6 (`Pending Cancel`) acknowledging the request.  The next would usually be 150=4 (`Canceled`) but there are corner cases.  The order could have been filled before the cancel was processed, or it could have ended for another reason.  So there's a few options here:
- Wait for the exec report to return any response with an understanding that the order might not have been cancelled (this implementation)
- Await the cancel to be fully processed, handling all expected responses (slowest and most expensive to maintain)
- Don't bother waiting for a response (fastest yet least deterministic)

Just want to ensure the behavior is well-understood.

MikeHathaway: It looks like it's only sending one response, with the switch being on `35=8` or `35=9`. I'm not seeing any `150=6` responses, unless this is getting swallowed?

Here's an example of a malformed request response: ```client received message 8=FIX.4.4|9=244|35=9|49=ERISX|56=ENRT637abe14c|34=11|52=20200515-20:40:24.721|37=281474978989673|11=bdb58658-f0a1-420f-8ad9-0fa99019e0$
4|41=a86354b1-3cc5-4b45-b4c1-9bfced7fdab5|39=8|60=20200515-20:40:24.702140697|434=1|102=1|58=UNKNOWN ORDER - 281474978989673|10=066```

And a successful request response: ```
8=FIX.4.4|9=343|35=8|49=ERISX|56=ENRT637abe14c|34=10|52=20200515-20:44:53.648|37=281474978990417|11=23363868-3907-4892-85bf-265f791d8e9
9|41=1de3e1fd-aede-4c1e-bcd1-8068e72f9d4d|17=281474987349296|150=4|39=4|1=ENR637abe14|55=ETH/USD|54=1|38=0.2|40=2|44=190.1|15=ETH|59=1|18=6|151=0|14=0|6=0|60=20200515-20:44:53.628614809|13=3|479=USD|58=USER INITIATED|7012=0|10=042
```
In this case, I believe this would be deterministic as it blocks until it receives a success or failure response.

EdNoepel: A 35=9 is a cancel reject response, and isn't expected to provide tag 150.  The 102=1 means the order wasn't in their book.  I'd expect a 102=0 or 102=3 if the order couldn't be cancelled for a legitimate reason.

Recommend discussing this with ErisX.

EdNoepel: Per our discussion, it seems ErisX is sending 102=1 instead of 102=0 in response to an order being filled by the time we submit our cancel.  This is nonstandard behavior, but a separate issue.

For the purposes of this review, I believe we should wait for either a 35=8 (success) or 35=9 (failure) response, and then log tags 102 and 58 if the cancel failed.

MikeHathaway: Updated accordingly

PR #137, opened 2020-05-16
Add coinone

MikeHathaway:


pyexchange/coinone.py
```diff
@@ -0,0 +1,256 @@
...
+        elif 10000 <= price < 50000:
+            return 10
+        elif 50000 <= price < 100000:
+            return 50
+        elif 100000 <= price < 500000:
+            return 100
+        elif 500000 <= price < 1000000:
+            return 500
+        elif 1000000 <= price:
+            return 1000
```
EdNoepel, line 118: Should we throw a runtime exception here if price is under 1000 or over 1000000?

pyexchange/coinone.py
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        result = self._http_authenticated_request("POST", f"/v2/order/cancel", data)
+        return True if result["result"] == "success" else False
+
+    def get_trades(self, pair: str, offset: int = 0) -> List[Trade]:
+        assert (isinstance(pair, str))
+        assert (isinstance(offset, int))
+
+        currency = pair.split('-')[0]
```
EdNoepel, line 179: Not sure I understand this.  Why would we only request trades for the quote token?  Wouldn't that include pairs for other base currencies?

pyexchange/coinone.py
```diff
@@ -0,0 +1,256 @@
...
+
+        result = self._http_authenticated_request("POST", f"/v2/order/complete_orders", {"currency": currency})
+        return list(map(lambda item: CoinoneTrade.from_message(item, pair), result['completeOrders']))
+
+    def _choose_nonce(self) -> int:
+        with self.last_nonce_lock:
+            time.sleep(0.1)
```
EdNoepel, line 186: What does the 100ms wait accomplish?

pyexchange/coinone.py
```diff
@@ -0,0 +1,256 @@
...
+        url = f"{self.api_server}{resource}"
+
+        return self._result(requests.request(method=method,
+                                             url=url,
+                                             data=data,
+                                             timeout=self.timeout))
+
+    def _result(self, result) -> Optional[dict]:
+        if not result.ok:
+            raise Exception(f"Coinone API invalid HTTP response: {http_response_summary(result)}")
```
EdNoepel, line 248: Recommend not raising base exception types; `RuntimeError` may work better here and on line 253 below.

pyexchange/coinone.py
MikeHathaway: Good point, adding the remaining price ranges now

pyexchange/coinone.py
MikeHathaway: Nothing, will remove it

pyexchange/coinone.py
MikeHathaway: The quote currency is KRW for every book, so their API just takes the base currency ETH or DAI, for example, and returns the trades for that currency assumed to be quoted in KRW.

pyexchange/coinone.py
MikeHathaway: Good point; updated accordingly.

pyexchange/coinone.py
```diff
@@ -245,12 +254,12 @@ def _http_unauthenticated_request(self, method: str, resource: str, body: dict):
...
     def _result(self, result) -> Optional[dict]:
         if not result.ok:
-            raise Exception(f"Coinone API invalid HTTP response: {http_response_summary(result)}")
+            raise RuntimeError(f"Coinone API invalid HTTP response: {http_response_summary(result)}")
 
         try:
             data = result.json()
-        except Exception:
-            raise Exception(f"Coinone API invalid JSON response: {http_response_summary(result)}")
+        except RuntimeError:
```
EdNoepel, line 261: You're more likely to get a `JSONDecodeError` error here, which subclasses `ValueError`.  Recommend catching `ValueError`.

ith-harvey: +1

PR #138, opened 2020-05-19
Update erisx with logic to handle mutual tls for production

MikeHathaway:


pyexchange/erisx.py
```diff
@@ -81,16 +81,25 @@ def __init__(self, fix_trading_endpoint: str, fix_trading_user: str,
         assert(isinstance(clearing_url, str) or (clearing_url is None))
         assert(isinstance(api_key, str) or (api_key is None))
         assert(isinstance(api_secret, str) or (api_secret is None))
+        assert(isinstance(certs, dict) or (certs is None))
         assert(isinstance(account_id, int))
 
         # enable access from sync_trades and inventory_service without overriding socket
         if fix_trading_endpoint is not None and fix_trading_user is not None:
-            self.fix_trading = ErisxFix(fix_trading_endpoint, fix_trading_user, fix_trading_user, password)
+            if certs is None:
```
EdNoepel, line 89: The switch seems to add an extra 6 lines of code.  You could just pass `certs` all the time, even if it's `None`.

pyexchange/erisx.py
```diff
@@ -72,7 +72,7 @@ class ErisxApi(PyexAPI):
 
     def __init__(self, fix_trading_endpoint: str, fix_trading_user: str,
                  fix_marketdata_endpoint: str, fix_marketdata_user: str, password: str,
-                 clearing_url: str, api_key: str, api_secret: str, account_id: int = 0):
+                 clearing_url: str, api_key: str, api_secret: str, certs: dict, account_id: int = 0):
```
EdNoepel, line 75: nit: `ErisxFix` class defaults the param to `None`; curious why that isn't done here.

tests/test_erisx.py
```diff
@@ -49,11 +49,12 @@ def setup_method(self):
         orig_get_account = ErisxApi.get_account
         FixEngine.logon = do_nothing
         ErisxApi.get_account = do_nothing
+        self.account_id = 0
         self.client = ErisxApi(fix_trading_endpoint="127.0.0.1:1752", fix_trading_user="test",
                                fix_marketdata_endpoint="127.0.0.1:1753", fix_marketdata_user="test",
                                password="test",
                                clearing_url="https://127.0.0.1/api/v1/",
-                               api_key="key", api_secret="secret", account_id=0)
+                               api_key="key", api_secret="secret", certs={}, account_id=self.account_id)
```
EdNoepel, line 57: Is the behavior passing an empty dictionary the same as passing `None`?

pyexchange/erisx.py
MikeHathaway: No good reason, updated accordingly.

pyexchange/erisx.py
MikeHathaway: Good point, updated accordingly.

tests/test_erisx.py
MikeHathaway: Functionally yes, but would cause issues if FIX tests were ever added. Replaced with `None`

pyexchange/erisx.py
```diff
@@ -400,3 +403,23 @@ def parse_orders_list(messages: List[simplefix.FixMessage]) -> List:
             orders.append(order)
 
         return orders
+
+    @staticmethod
+    def parse_order_book(message: simplefix.FixMessage) -> List:
+        trade_count = int(message.get(268))
```
EdNoepel, line 409: Regarding nomenclature:  When I think of a level 2 _order book_, I think of two lists of _quotes_; a list of buys and a list of sells, each sorted by price.  The name `trade` implies an execution.

pyexchange/erisx.py
```diff
@@ -223,30 +230,13 @@ def get_trades(self, pair: str, page_number: int = 8) -> List[Trade]:
...
-        return True if response else False
+    # TODO: Not currently available
+    def get_all_trades(self, pair, page_number) -> List[Trade]:
+       pass
 
-    # TODO: finish implementing
-    def get_all_trades(self, pair: str, page_number: int = 1) -> List[Trade]:
-        assert(isinstance(pair, str))
-        assert(isinstance(page_number, int))
+    def get_orderbook(self, pair: str, page_number: int = 1) -> List[Trade]:
```
EdNoepel, line 237: I don't believe we should be creating `Trade` objects for this.  Other exchange implementations seem to return raw data from the exchange API.  Please see other comment in `ErisxFix.parse_order_book`.

pyexchange/erisx.py
MikeHathaway: Agreed, that makes more sense. Updated accordingly.

pyexchange/erisx.py
MikeHathaway: Updated typing accordingly

PR #140, opened 2020-05-21
add support for automated deposits and withdrawals for dydx

MikeHathaway:


pyexchange/dydx.py
```diff
@@ -144,26 +150,68 @@ def get_orders(self, pair: str) -> List[Order]:
...
         open_orders = filter(lambda order: order['status'] == 'OPEN', orders['orders'])
-        
+
         market_info = self.market_info[pair]
 
         return list(map(lambda item: DydxOrder.from_message(item, pair, market_info), open_orders))
 
-    def deposit_funds(self, token, amount: float):
-        assert (isinstance(amount, float))
+    def _set_allowances(self) -> True:
```
MikeHathaway, line 158: Not a huge fan of this approach, but it errors out if it's called more than once, and has to be executed prior to any deposits.

MikeHathaway: I could call their contract directly, as they do, here: https://github.com/dydxprotocol/dydx-python/blob/4dc8df8228cc0fa8d52b810209ea116281be8409/dydx/eth.py#L155

However, this is a non public API and may break unexpectedly in the future.

EdNoepel: This seems a good example of there being no "right way" to do it.  If I were implementing this, I would have left this method public (no leading underscore), and omitted the `is_new` in the constructor.  I'd document the `deposit_funds` to warn the user to call `set_allowances` before making the first deposit.  If possible, I'd catch the error thrown upon a failed deposit and rethrow it with an error asking them to check allowances.

MikeHathaway: Thanks for the suggestion. That seems like a much better approach. I've adjusted accordingly.

PR #141, opened 2020-05-27
don't attempt to verify self signed certs in erisx

MikeHathaway:


MikeHathaway:
> If this engine is eventually used for other exchanges, we should consider making the verify mode configurable.

Yeah that makes sense - although I would assume that we would still be using self-signed certificates for other exchanges?

EdNoepel:
> > If this engine is eventually used for other exchanges, we should consider making the verify mode configurable.
> 
> Yeah that makes sense - although I would assume that we would still be using self-signed certificates for other exchanges (given we lack a common CA)?

Some exchanges/consumers might want to use `letsencrypt` or pay for a common CA.

PR #146, opened 2020-06-09
convert bitso cancel result stringified bool to bool

MikeHathaway:


pyexchange/bitso.py
```diff
@@ -162,7 +162,7 @@ def cancel_order(self, order_id: str) -> bool:
         self.logger.info(f"Cancelling order #{order_id}...")
 
         result = self._http_authenticated_request("DELETE", f"/v3/orders/{order_id}", {})
-        return result["success"]
+        return True if result["success"] == "True" else False
```
EdNoepel, line 165: Methinks `return result["success"] == "True"` would've been sufficient.

MikeHathaway: Good point, I simplified accordingly.

PR #149, opened 2020-07-14
temporarily update fix.py logging levels

MikeHathaway:


pyexchange/fix.py
```diff
@@ -288,7 +288,7 @@ async def _heartbeat(self):
                 m = self.create_message(simplefix.MSGTYPE_HEARTBEAT)
                 await self._write_message(m)
             except ConnectionError as ex:
-                logging.warning(f"Unable to send heartbeat: {ex}")
+                logging.info(f"Unable to send heartbeat: {ex}")
```
EdNoepel, line 291: This should remain a `warning` because it's unexpected behavior.

MikeHathaway:
> Is there a reason the keeper cannot be run with the `--debug` switch in production to show this data?

The idea was to avoid creating excessively large log files with debug information for modules that we aren't specifically interested in. So would have to make another PR to revert these changes once we've established a solution

pyexchange/fix.py
MikeHathaway: True; was being overly cautious to ensure it logs out but I can change this back

PR #150, opened 2020-07-15
Add logging to ErisX connection logic; Change timeinforce to cancel on disconnect

MikeHathaway:


pyexchange/fix.py
```diff
@@ -258,16 +258,19 @@ def _run_session(self):
...
+                self.ssl_context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+                self.ssl_context.load_cert_chain(certfile=self.certs['client_cert'], keyfile=self.certs['client_key'])
+                self.ssl_context.check_hostname = False
+                self.ssl_context.verify_mode = ssl.CERT_NONE
+                self.reader, self.writer = await asyncio.open_connection(address, port, loop=self.session_loop, ssl=self.ssl_context)
+            else:
+                self.reader, self.writer = await asyncio.open_connection(address, port, loop=self.session_loop)
+
+            self.connection_state = FixConnectionState.CONNECTED
+        except Exception as e:
```
MikeHathaway, line 272: Not happy with using base exception; but i'm not yet sure what type of exception could be thrown

pyexchange/erisx.py
```diff
@@ -184,7 +184,6 @@ def place_order(self, pair: str, is_sell: bool, price: float, amount: float) ->
         message.append_pair(simplefix.TAG_ORDERQTY, amount)
         message.append_pair(simplefix.TAG_ORDTYPE, simplefix.ORDTYPE_LIMIT)  # always place limit orders
         message.append_pair(simplefix.TAG_PRICE, price)
-        message.append_pair(simplefix.TAG_TIMEINFORCE, simplefix.TIMEINFORCE_GOOD_TILL_CANCEL)
```
EdNoepel, line 187: I confirmed tag 59 is optional.

pyexchange/fix.py
EdNoepel: Fair enough.  And sometimes it's acceptable to put a catch-all at the root level of a thread.

PR #151, opened 2020-07-17
Add uniswapv2

MikeHathaway:


pyexchange/graph.py
```diff
@@ -0,0 +1,172 @@
...
+import time
+import logging
+import requests
+
+from typing import Optional
+from json import JSONDecodeError
+from lib.pymaker.pymaker.util import http_response_summary
+
+get_balances = '''
+  # liquidityPositions(where: {user: {id: "0x332f60EDC783E4Db3E0a18F8dFEB368Ae178CCd9"}}) {
```
EdNoepel, line 27: What is this the address of?  Should it be parameterized?

pyexchange/uniswapv2.py
```diff
@@ -0,0 +1,342 @@
...
+                 # ec_signature_r: Optional[str], ec_signature_s: Optional[str], ec_signature_v: Optional[int]):
+        assert (isinstance(web3, Web3))
+        assert (isinstance(token_a, Token))
+        assert (isinstance(token_b, Token))
+
+        self.web3 = web3
+        self.token_a = token_a
+        self.token_b = token_b
+        self.router_address = Address("0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D")
+        self.factory_address = Address("0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f")
```
EdNoepel, line 55: Hardcoding these addresses prevents testing outside of _mainnet_.  Also the **router** address is not immutable; that contract can be swapped out because no balances are stored there.

pyexchange/uniswapv2.py
```diff
@@ -0,0 +1,342 @@
...
+        return Wad(self._pair_contract.functions.MINIMUM_LIQUIDITY(self.account_address.address).call())
+
+    def get_pair_address(self, token_a_address: Address, token_b_address: Address) -> Address:
+        assert (isinstance(token_a_address, Address))
+        assert (isinstance(token_b_address, Address))
+
+        return Address(self._factory_contract.functions.getPair(token_a_address.address, token_b_address.address).call())
+
+    # TODO: determine appropriate amount default
+    def approve(self, token: Token, amount: int = 10):
```
EdNoepel, line 118: Do we need a default?  I don't see an interface contract which mandates omission of the parameter.

pyexchange/uniswapv2.py
```diff
@@ -0,0 +1,342 @@
...
+        assert (isinstance(token, Token))
+        assert (isinstance(amount, int))
+
+        erc20_token = ERC20Token(self.web3, token.address)
+
+        approval_function = directly()
+        return approval_function(erc20_token, self.router_address, 'IUniswapV2Router02')
+
+    def get_block(self) -> Transact:
```
EdNoepel, line 127: What's the purpose of this wrapper function?

pyexchange/uniswapv2.py
```diff
@@ -0,0 +1,342 @@
...
+            self.account_address.address,
+            self._deadline()
+        ]
+
+        return Transact(self, self.web3, self.router_abi['abi'], self.router_address, self._router_contract,
+                        'addLiquidityETH', addLiquidityArgs, {'value': amounts['amount_eth_desired'].value})
+
+    # TODO: finish implementing
+    # Enable liquidity to be removed from a pool up to a set limit
+    def permit_removal(self, pair, amount: Wad) -> Transact:
```
EdNoepel, line 224: There's not a lot of value keeping this prototype around.  If there was some implementation behind it, I'd recommend commenting it out.  As is, seems best to remove it.

pyexchange/uniswapv2.py
```diff
@@ -0,0 +1,342 @@
...
+            amounts['amountAMin'].value,
+            amounts['amountBMin'].value,
+            self.account_address.address,
+            self._deadline()
+        ]
+
+        return Transact(self, self.web3, self.router_abi['abi'], self.router_address, self._router_contract,
+                        'removeLiquidity', removeLiquidityArgs)
+
+    # TODO: add switch to handle whether or not a givne pool charges a fee
```
EdNoepel, line 254: typo `given`

tests/test_uniswapv2.py
```diff
@@ -0,0 +1,148 @@
...
+        # when
+        # self._router_contract.approve([self.dai], directly())
+        approval_function = directly()
+        
+        approval_function(self.dai, self.router_address, 'IUniswapV2Router02')
+
+        # then
+        assert self.dai.allowance_of(self.our_address, self.router_address) > Wad(0)
+
+    def test_getting_balances(self):
```
EdNoepel, line 89: As written, this test merely ensures no exception is raised.

tests/test_uniswapv2.py
```diff
@@ -0,0 +1,148 @@
...
+        approval_function(self.dai, self.router_address, 'IUniswapV2Router02')
+
+        time.sleep(20)
+        # when
+        add_liquidity = self.uniswap.add_liquidity_eth(eth_pair_amounts, self.dai.address).transact()
+
+        # then
+        assert add_liquidity.result == True
+
+        # when
```
EdNoepel, line 119: Test seems incomplete

tests/test_uniswapv2.py
```diff
@@ -0,0 +1,148 @@
...
+        orders = self.uniswap.get_orders(self.pair, [zrx_order])
+        # then
+        assert orders[0].order_id is not None
+        assert orders[0].is_sell is False
+        assert orders[0].price == Wad.from_number(45.0)
+        assert orders[0].amount == Wad.from_number(5.0)
+        assert orders[0].zrx_order == zrx_order
+
+    def test_swap():
```
EdNoepel, line 147: Test is empty, and missing newline at end of file

pyexchange/uniswapv2.py
MikeHathaway: The address is the same across all networks, but that being said, I agree hardcoding is non ideal. Do you have a preferred solution to this? I've seen both Enums and address config files in the rest of the codebase.

pyexchange/uniswapv2.py
MikeHathaway: Was used to debug web3 connectivity, but I've since removed it.

pyexchange/graph.py
MikeHathaway: This was for testing balance retrieval. It was parameterized elsewhere, so I removed all of this old testing code.

pyexchange/uniswapv2.py
EdNoepel: I'm happy keeping them as ctor arguments with default values, as you've implemented.  Keepers can override them with commandline arguments if desired.

tests/test_uniswapv2.py
MikeHathaway: Added additional working tests. That being said, the tests currently only work with a local ganache instance. I am working through getting it working the docker-compose managed parity node.

pyexchange/uniswapv2.py
MikeHathaway: Not really, I was thinking about passing it through as a `kwarg`, but that doesn't seem to be necessary / will potentially add too much complexity. I've removed it.

tests/test_uniswapv2.py
MikeHathaway: Updated test

tests/test_uniswapv2.py
MikeHathaway: Updated tests

tests/test_uniswapv2.py
MikeHathaway: Updated tests

pyexchange/uniswapv2.py
MikeHathaway: Agreed, I've removed it

pyexchange/uniswapv2.py
MikeHathaway: Removed TODO entirely

pyexchange/uniswapv2.py
MikeHathaway: Left as is, per discussion.

MikeHathaway:
> Most recent push looks good. Still curious about the `dcoker-compose.yml`. Seems verbatim with the `pymaker` one, which could be reused. Does `test.sh` really need both ganache and the dockerized testchain?

No, it doesn't, and thank you for catching that. I pushed up a commit that removes references to the pymaker image.

PR #154, opened 2020-08-07
update dydx client version and set to spot trading only

MikeHathaway:


MikeHathaway:
This requires more substantial changes to support balance retrieval.

PR #155, opened 2020-08-07
Return empty list if Bittrex result is None

grandizzy:


grandizzy:
> I suppose `self._result` method would produce an empty dict on line 287, which would be falsey here.

not an empty dict but rather a dict like `{"success":false,"message":"INVALID_MARKET","result":null}`

PR #156, opened 2020-08-07
 Coinbase transfers between account and subaccounts

EdNoepel:
Implemented API to get profiles and transfer between default account and subaccounts.  Note the `_result` had to be reworked because the POST to `/profiles/transfer` doesn't have a JSON response.

### DIFF
--- a/pyexchange/coinbase.py
+++ b/pyexchange/coinbase.py
@@ -243,6 +243,45 @@ class CoinbaseApi(PyexAPI):
 
         return list(map(lambda item: Trade.from_all_list(pair, item), result))
 
+    def get_profiles(self):
+        # Returns list of profiles; one profile exists for each portfolio/account/subaccount
+        return self._http_authenticated("GET", "/profiles", {})
+
+    def get_profile(self, portfolio: str):
+        # Retrieves the profile identifier for a particular account or subaccount based on portfolio name
+        assert isinstance(portfolio, str)
+
+        matching = list(filter(lambda p: p['name'] == portfolio, self.get_profiles()))
+        if len(matching) > 1:
+            raise ValueError("Multiple profiles match that portfolio name; please adjust names via Coinbase Pro GUI")
+        elif len(matching) < 1:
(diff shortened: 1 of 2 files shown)


pyexchange/coinbase.py
```diff
@@ -318,13 +357,15 @@ def _http_unauthenticated(self, method: str, resource: str, body: dict):
...
-
-        try:
-            data = result.json()
-        except Exception:
-            raise Exception(f"Coinbase API invalid JSON response: {http_response_summary(result)}")
-
-        return data
+            raise RuntimeError(f"Coinbase API response: {http_response_summary(result)}")
+
+        if result.content and result.content != b'OK':
```
ith-harvey, line 365: Hey Ed, could you explain why you're using the byte literal here? Is this the value stored in `result.content`? I assume this is some of the "reworking" you were referring to

tests/manual_test_coinbase.py
```diff
@@ -15,14 +15,19 @@
...
 from pyexchange.coinbase import CoinbaseApi
 from pymaker import Address, Wad
 
+logging.basicConfig(format='%(asctime)-15s %(levelname)-8s %(message)s', level=logging.INFO)
+logging.getLogger('urllib3').setLevel(logging.INFO)
+logging.getLogger('urllib3.connectionpool').setLevel(logging.INFO)
+logging.getLogger("requests").setLevel(logging.INFO)
+logging.getLogger('requests.packages.urllib3.connectionpool').setLevel(logging.INFO)
 
+# ctor params: api_server: str, api_key: str, secret_key: str, password: str
```
ith-harvey, line 30: not quite sure what `ctor` is referring to?

pyexchange/coinbase.py
EdNoepel: Yes, that's the success response when Coinbase doesn't return JSON.

tests/manual_test_coinbase.py
EdNoepel: Short for "constructor".  It's the parameters to pass.

PR #157, opened 2020-08-20
Binance.us API wrapper

Exef:
RestAPI documentiation: https://github.com/binance-us/binance-official-api-docs/blob/master/rest-api.md

Questions:
- [x] To cancel an order it's required to provide its pair. I thought about implementing a mapping of orders to pairs, but I couldn't find any mention that `orderId` for all pairs. Who should manage this mapping - API wrapper or MM-Keeper?
- [x] I'm not sure if I should use float conversion on `Wad`.
- [x] What kind of order should be used to place in the `place_order` method? I placed - `Good Till Cancel`, but other options are: `Immediate Or Cancel` and  `Fill or Kill` 
- [x] Should I implement rate limiters handling in the `_result` method of the API wrapper?


pyexchange/binance_us.py
```diff
@@ -0,0 +1,267 @@
...
+    def create(item):
+        return Order(order_id=item['orderId'],
+                     pair=item['symbol'],
+                     is_sell=True if item['side'] == 'SELL' else False,
+                     price=Wad.from_number(item['price']),
+                     amount=Wad.from_number(item['origQty']))
+
+
+class Trade:
+    def __init__(self,
```
MikeHathaway, line 84: Instead of redefining a `Trade` and `Order` class, you can also just import the standard `Trade` and `Order` classes from `pyexchange.model`, and utilize the subclass builder pattern to instantiate a new Order and Trade. You can see an example here: https://github.com/makerdao/pyexchange/blob/master/pyexchange/dydx.py#L52

Exef: Awesome, I'll use it.

MikeHathaway:
BinanceUsApi class looks great! I made a comment about subclassing Order and Trade classes, but otherwise looks good. I would also suggest adding some mock based unit tests. Examples can be seen in `tests/test_okex.py` or `tests/test_dydx.py`

MikeHathaway:
> RestAPI documentiation: https://github.com/binance-us/binance-official-api-docs/blob/master/rest-api.md
> 
> Questions:
> 
>     * [ ]  To cancel an order it's required to provide its pair. I thought about implementing a mapping of orders to pairs, but I couldn't find any mention that `orderId` for all pairs. Who should manage this mapping - API wrapper or MM-Keeper?

I've been handling similar situations in MM-Keeper. I simply overrode the `cancel_order` signature in pyexchange, and then when I register that function in MM-Keeper, I pass any necessary args. The ErisX exchange integration had the same issue with cancel orders requiring an `orderId`.

> 
>     * [ ]  I'm not sure if I should use float conversion on `Wad`.

Other people may have different opinions, but I would recommend primarily operating on `Wad`, and then just converting right before you message the exchange. MM-Keeper is relatively more complicated, and all operations are conducted in Wad there, so i've found it helpful to maintain consistency.

> 
>     * [ ]  What kind of order should be used to place in the `place_order` method? I placed - `Good Till Cancel`, but other options are: `Immediate Or Cancel` and  `Fill or Kill`
> 
>     * [ ]  Should I implement rate limiters handling in the `_result` method of the API wrapper?
Probably not, `_result` is IMO more about handling error responses and standardizing the formatting. There's logic in MM-Keeper `synchronize_orders()` that checks balances, orders, bands, etc and decides whether or not to place or cancel orders. It might be a better fit there

PR #161, opened 2020-08-27
Migrate Bittrex integration to API v3

Exef:
TODOs: 
- [x] Gather real data for mocked tests 

Points to double-check:
1. In the [upgrade guide](https://bittrex.github.io/guides/v3/upgrade) it stated that `Market symbols are reversed in v3 to align with standard forex conventions`. I need to clarify if it's needed to change the order of the provided `pair`.
2. In the V3 `QuantityRemaining` is replaced with `fillQuantity` I replaced it in `Order` and used it in `get_trades` `Trade` constructor in place of calculation `item['Quantity'] - item['QuantityRemaining']))`
3. `PricePerUnit` field is depracated in `get_trades` API response I replaced it with `limit`


pyexchange/bittrex.py
```diff
@@ -174,25 +93,21 @@ def place_order(self, pair: str, is_sell: bool, price: Wad, amount: Wad) -> str:
...
 
-        params = {
-            "quantity": str(amount),
-            "rate": str(price),
-            "market": pair
+        body = {
+            'marketSymbol': pair,
+            'direction': "SELL" if is_sell else "BUY",
+            'quantity': float(price),
+            'limit': float(price),
```
ith-harvey, line 100: did you experiment turning this into a string rather than a float? float can cause some decimal issues.

ith-harvey (commented):
Curious to hear a response about floats...

pyexchange/bittrex.py
```diff
@@ -28,110 +29,28 @@
 from urllib.parse import urlencode
 
 from pyexchange.api import PyexAPI
+from pyexchange.model import Order as BaseOrder, Trade
```
MikeHathaway, line 32: BaseOrder doesn't seem to be defined here?

pyexchange/bittrex.py
```diff
@@ -148,23 +67,23 @@ def __init__(self, api_server: str, api_key: str, secret_key: str, timeout: floa
...
+    def get_precision(self, pair: str):
+        assert(isinstance(pair, str))
+        return self._http_request("GET", f"/v3/markets/{pair}", {})["precision"]
 
     def get_pair(self, pair: str):
         assert(isinstance(pair, str))
-        return next(filter(lambda symbol: symbol['MarketName'] == pair, self.get_markets()))
+        return next(filter(lambda market: market['symbol'] == pair, self.get_markets()))
 
     def get_balances(self):
```
MikeHathaway, line 80: Minor point, but can you add the return Types for these functions? Helps with jumping into the codebase later

MikeHathaway: Oh awesome find! Yeah that would be a great addition.

MikeHathaway (approved):
LGTM!

PR #164, opened 2020-08-31
Dydx spot migration and client update

MikeHathaway:


EdNoepel:
> > Changes look good. Please investigate build failure. Seems to be an issue with the ganache docker image as well as an issue with dydx test setup.
> 
> Seems to have been a random failure due to the container not initializing in time. I haven't been able to repro the failure locally, and when I triggered a new build with a minor manual test update, everything passed

That happens in `auction-keeper` too; no worries.  I usually wind up tweaking the `README.md` to fix it.  :innocent:

PR #165, opened 2020-09-03
small new method to get tickSize

mitakash:
small method to help new Market Maker keeper for leverj futures that uses DAI as a collateral


ith-harvey (approved):
LGTM

PR #168, opened 2020-09-10
some new methods added to be used in market maker keeper

mitakash:


pyexchange/leverjfutures.py
```diff
@@ -239,6 +244,12 @@ def get_tickSize(self, pair: str):
         assert(isinstance(pair, str))
         return self.get_product(pair)["tickSize"]
 
+    def get_minimum_order_quantity(self, pair: str):
```
ith-harvey, line 247: could you add a comment in this function describing what it does?

mitakash: yup adding it

ith-harvey (approved):

PR #169, opened 2020-09-19
added new functions to deal with Leverj Futures id

mitakash:


pyexchange/leverjfutures.py
```diff
@@ -127,7 +127,7 @@ def __repr__(self):
 
     @staticmethod
     def from_our_list(pair, trade):
-        return Trade(trade_id=trade['executionid'],
+        return Trade(trade_id=trade['executionId'],
```
MikeHathaway, line 130: More of a Quality of life change, but it might be worth subclassing the `Trade` object as a `LeverJFutureTrade; likewise with `Order`. An example of this can be seen in: https://github.com/makerdao/pyexchange/blob/master/pyexchange/dydx.py#L53. This would enable us to cut out ~100 lines of code, and have a single source of truth for these Classes.

pyexchange/leverjfutures.py
```diff
@@ -192,6 +191,11 @@ def get_quote_balance(self, quote_asset_address: str) -> str:
         quote_balance = balances[quote_asset_address]['available']
         return quote_balance
 
+    def get_plasma_balance(self, quote_asset_address: str) -> str:
```
MikeHathaway, line 194: Might want to use `Address` type instead of `str` as this provides additional safety. Also, it would be good to an `assert(isinstance(quote_asset_address, <TYPE>)` for clearer error messages.

pyexchange/leverjfutures.py
```diff
@@ -267,27 +271,37 @@ def get_orders(self, pair: str) -> List[Order]:
...
         return list(map(lambda item: Order.from_list(item, pair), result_pair))
 
+    def get_id_from_pair(self, pair: str) -> str:
+        assert(isinstance(pair, str))
+        if pair=="BTCUSD":
+            return "1"
+        elif pair=="ETHUSD":
+            return "2"
+        else:
+            self.logger.info(f'You have passed in an unsupported pair')
```
MikeHathaway, line 281: What happens if `None` is returned?

pyexchange/leverjfutures.py
```diff
@@ -267,27 +271,37 @@ def get_orders(self, pair: str) -> List[Order]:
...
         assert(isinstance(page_number, int))
-        result = self._http_authenticated("GET", "/futures/api/v1", f"/instrument/{pair}/trade", None)
+
+        result = self._http_authenticated("GET", "/futures/api/v1", f"/instrument/{self.get_id_from_pair(pair)}/trade", None)
 
         return list(map(lambda item: Trade.from_all_list(pair, item), result))
 
     def get_symbol_trades(self, symbol: str):
-        return self._http_authenticated("GET", "/futures/api/v1", f"/instrument/{symbol}/trade", None)
+        return self._http_authenticated("GET", "/futures/api/v1", f"/instrument/{get_id_from_pair(symbol)}/trade", None)
```
MikeHathaway, line 304: Should this be `self.get_id_from_pair()`?

pyexchange/leverjfutures.py
mitakash: Well in this specific section Leverj spot is using different fields that Leverj futures exchange, but yes i tend to agree with you. Let's leave that for version 2.

MikeHathaway: Fair point; sounds good.

MikeHathaway (approved):
LGTM!

PR #171, opened 2020-09-23
Uniswap trade sync

MikeHathaway:


pyexchange/uniswapv2_analytics.py
```diff
@@ -33,57 +34,100 @@
...
         assert (isinstance(pair, str))
-        assert (isinstance(token_a, Token))
-        assert (isinstance(token_b, Token))
+        assert (isinstance(base_token, Token))
+        assert (isinstance(our_liquidity_balance, Wad))
+        assert (isinstance(previous_base_token_reserves, Wad))
 
-        if trade['pair']['token0']['id'] == token_a.address.address:
-            swap_price = Wad.from_number(trade['pair']['token1Price'])
+        our_pool_share = our_liquidity_balance / Wad.from_number(trade['pair']['totalSupply'])
```
EdNoepel, line 57: What happens if this is called when we have no liquidity in the pool?

EdNoepel (commented):
If I understand correctly, this will synthesize a single trade for each hour liquidity was removed from the pool, based upon our supply of the token in that pool.  As such, a consumer would be expected to query this once per hour.  Is my understanding correct?

pyexchange/uniswapv2_analytics.py
MikeHathaway: > If I understand correctly, this will synthesize a single trade for each hour liquidity was removed from the pool, based upon our supply of the token in that pool. As such, a consumer would be expected to query this once per hour. Is my understanding correct?

Yes, that is our current plan. This would be enforced by setting the `refresh-frequency` in `sync-trades`.

MikeHathaway: If a user has no liquidity in the pool than it will return `[]`, based on the logic on line 277. If it would be better, I can also add an assertion to `from_our_trades_message` that will ensure the incoming `our_liquidity_balance` is greater than 0.

EdNoepel (approved):

PR #172, opened 2020-09-25
add listener for erisx cancelled messages

MikeHathaway:


pyexchange/fix.py
```diff
@@ -40,6 +40,7 @@ def fprint(encoded_msg):
     return encoded_msg.replace(b"\x01", b"|")
 
 
+# TODO: add callback based absctration that can be subclassed for various socket listeners
```
EdNoepel, line 43: nit: spelling `abstraction`

pyexchange/fix.py
```diff
@@ -208,6 +209,41 @@ def wait_for_get_orders_response(self) -> List[simplefix.FixMessage]:
...
+                message = self.application_messages.get()
+                assert isinstance(message, simplefix.FixMessage)
+
+                # handle message cancellation
+                if message.get(simplefix.TAG_MSGTYPE) == simplefix.MSGTYPE_EXECUTION_REPORT:
+
+                    if message.get(simplefix.TAG_EXECTYPE) in cancel_message_types:
+                        order_to_cancel_id = f"{message.get(simplefix.TAG_ORDERID)}|{message.get(simplefix.TAG_CLORDID)}"
+
+                        if message.get(b'5001') == '4'.encode('utf-8'):
```
EdNoepel, line 232: This tag is likely specific to ErisX.  If we add support for another exchange, we'll need to move this logic into `ErisxFix` class.

EdNoepel (approved):
Some comments within, looks okay.

pyexchange/fix.py
MikeHathaway: That makes sense to me, I was also thinking about moving all of these `wait_for_response` methods to the ErisXFix class in erisx.py entirely since the logic is largely exchange specific.

pyexchange/fix.py
MikeHathaway: Removed TODO entirely

PR #173, opened 2020-09-29
update uniswap analytics time range handling

MikeHathaway:


pyexchange/uniswapv2_analytics.py
```diff
@@ -266,24 +305,39 @@ def get_trades(self, pair: str, page_number: int = 1) -> List[Trade]:
...
         last_mint_event = mint_events[0]
+        last_burn_event = mint_events[0]
 
         our_liquidity_balance = Wad.from_number(last_mint_event['liquidity'])
         current_liquidity = self.get_current_liquidity(pair_address)
 
         last_mint_timestamp = int(last_mint_event['timestamp'])
+        last_burn_timestamp = int(last_burn_event['timestamp'])
         two_days_ago_unix = int(time.time() - (49 * 60 * 60))
+        mints_in_last_two_days = list(filter(lambda mint: int(mint['timestamp']) > two_days_ago_unix, mint_events))
```
EdNoepel, line 318: At some point in the future, it'll make sense to filter time in the query rather than querying years of history.  Not sure if _thegraph_ supports such a filter.

pyexchange/uniswapv2_analytics.py
```diff
@@ -266,24 +305,39 @@ def get_trades(self, pair: str, page_number: int = 1) -> List[Trade]:
...
+        elif current_liquidity != Wad.from_number(0) and len(mints_in_last_two_days) >= 1:
+            start_timestamp = mints_in_last_two_days[-1]['timestamp']
+            end_timestamp = current_time
+        elif current_liquidity != Wad.from_number(0) and len(mints_in_last_two_days) == 0:
             start_timestamp = two_days_ago_unix
+            end_timestamp = current_time
+        elif current_liquidity == Wad.from_number(0) and len(mints_in_last_two_days) >= 1:
+            start_timestamp = mints_in_last_two_days[-1]['timestamp']
+            end_timestamp = last_burn_timestamp
+        elif current_liquidity == Wad.from_number(0) and len(mints_in_last_two_days) == 0:
```
EdNoepel, line 333: Seems a bit redundant; the `else` below could cover this case.  Unless you just wanted it here for setting a breakpoint.

EdNoepel (approved):
LGTM

pyexchange/uniswapv2_analytics.py
MikeHathaway: Yeah that was the idea, but that would be much less confusing for future readers. I will update it accordingly.

pyexchange/uniswapv2_analytics.py
MikeHathaway: Yeah that's a good point - I had looked into it but, unfortunately, they don't really have a great time based filter available. The closest approximation would be block height - but that would increase complexity correspondingly, and I figured it was easier to just do it on the client given this should only be called once an hour or so.

MikeHathaway: Yeah that makes sense. We would be able to add logic around a starting timestamp / block, but I figured it wasn't worth the added complexity since we would only be making the call once an hour

PR #176, opened 2020-10-01
uniswapv2 syncer bugfixes

MikeHathaway:


pyexchange/uniswapv2_analytics.py
```diff
@@ -306,32 +306,35 @@ def get_trades(self, pair: str, page_number: int = 1) -> List[Trade]:
...
         burn_events = self.get_our_burn_txs(pair_address)
-        last_mint_event = mint_events[0]
-        last_burn_event = mint_events[0]
+
+        current_time = int(time.time())
+        two_days_ago_unix = int(time.time() - (49 * 60 * 60))
+
+        if len(mint_events) == 0:
+            return trades_list
+        else:
```
EdNoepel, line 315: The `else` confuses me a bit.  Isn't the rest of the method implied to be part of the `else` condition?  Seems it'd be more readable to keep it at the same indentation level, unless I'm misinterpreting.

MikeHathaway: Yes, good point. I've removed the else condition entirely, and move it's logic to the outer indentation level.

EdNoepel (approved):
👌

PR #177, opened 2020-10-05
use pairhourid as opposed to timestamp for generating trade id

MikeHathaway:
There are rounding issues that appear in the response objects price and amount attributes from Graph Protocol. This is causing some of the trades to be unnecessarily updated. To avoid excessively updating large numbers of trades, this PR stores the last time it queried and uses that timestamp as the starting point for any future queries.

### DIFF
--- a/pyexchange/uniswapv2_analytics.py
+++ b/pyexchange/uniswapv2_analytics.py
@@ -320,22 +320,13 @@ class UniswapV2Analytics(Contract):
         mints_in_last_two_days = list(filter(lambda mint: int(mint['timestamp']) > two_days_ago_unix, mint_events))
 
         if current_liquidity != Wad.from_number(0) and len(mints_in_last_two_days) >= 1:
-            if self.our_last_pair_hour_timestamp > mints_in_last_two_days[-1]['timestamp']:
-                start_timestamp = self.our_last_pair_hour_timestamp
-            else:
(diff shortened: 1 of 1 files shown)


pyexchange/uniswapv2_analytics.py
```diff
@@ -327,13 +320,22 @@ def get_trades(self, pair: str, page_number: int = 1) -> List[Trade]:
         mints_in_last_two_days = list(filter(lambda mint: int(mint['timestamp']) > two_days_ago_unix, mint_events))
 
         if current_liquidity != Wad.from_number(0) and len(mints_in_last_two_days) >= 1:
-            start_timestamp = mints_in_last_two_days[-1]['timestamp']
+            if self.our_last_pair_hour_timestamp > mints_in_last_two_days[-1]['timestamp']:
+                start_timestamp = self.our_last_pair_hour_timestamp
```
EdNoepel, line 324: For these three blocks, would it be more readable to use `max` function instead of a conditional?  For example, this could replace four lines of code:
```
start_timestamp = max(self.our_last_pair_hour_timestamp, mints_in_last_two_days[-1]['timestamp'])
```

MikeHathaway: Yeah, that would be much cleaner. I will update accordingly.

EdNoepel (approved):
LGTM

PR #179, opened 2020-10-07
Update uniswapv2_analytics get_trades around time traveling block based queries

MikeHathaway:


pyexchange/uniswapv2_analytics.py
```diff
@@ -150,6 +150,9 @@ def __init__(self, web3: Web3, token_config_path: str, keeper_address: Address,
             self._last_config = None
             self.token_config = self.get_token_config().token_config
 
+    def get_current_block(self) -> int:
+        return int(self.web3.eth.getBlock('latest')['number'])
```
EdNoepel, line 154: `web3.eth.blockNumber` is the usual way we do it in `pymaker`.

pyexchange/uniswapv2_analytics.py
```diff
@@ -388,53 +410,38 @@ def get_all_trades(self, pair: str, page_number: int = 1) -> List[Trade]:
...
-            if trade['pair']['token0']['id'] == base_token.address.address:
-                previous_base_token_reserves = Wad.from_number(pair_hour_data_list[0]['reserve0'])
-            else:
-                previous_base_token_reserves = Wad.from_number(pair_hour_data_list[index - 1]['reserve0'])
+                trades_list.append(UniswapTrade.from_all_trades_message(block_trade, pair, base_token, previous_base_token_reserves, timestamp))
 
-            formatted_trade = UniswapTrade.from_all_trades_message(trade, pair, base_token, previous_base_token_reserves)
-            trades_list.append(formatted_trade)
+            # assume there's 240 blocks in an hour
+            checked_block += 240
```
EdNoepel, line 439: On average there's 266.7 blocks per hour.  So we're checking 54 minutes of blocks here, which seems fine as an arbitrary way to batch the requests.

pyexchange/uniswapv2_analytics.py
```diff
@@ -388,53 +410,38 @@ def get_all_trades(self, pair: str, page_number: int = 1) -> List[Trade]:
 
         trades_list = []
 
-        two_days_ago_unix = int(time.time() - (48 * 60 * 60))
-        start_time = two_days_ago_unix if two_days_ago_unix > self.all_last_pair_hour_timestamp else self.all_last_pair_hour_timestamp
+        # calculate starting block, assuming there's 15 seconds in a given block
+        current_block = self.get_current_block()
+        two_day_ago_block = int(current_block - (4 * 60 * 25))
```
EdNoepel, line 415: Assuming a block is 13.5 seconds, this block would be 22.5 hours ago.
But the variable name is `two_day_ago_block`; did you forget to multiply by 2 here?  Or should the variable name be adjusted?

EdNoepel (commented):
Looks good, but I made a few comments.

pyexchange/uniswapv2_analytics.py
MikeHathaway: Is there an advantage to using one over another?

pyexchange/uniswapv2_analytics.py
MikeHathaway: Yes, good point - I will update the variable names.

pyexchange/uniswapv2_analytics.py
MikeHathaway: Yeah i'm not super enthused with the blocks/hour assumption, but I was afraid of making a separate set of calls for each period to get the associated timestamp - block relationship.

pyexchange/uniswapv2_analytics.py
```diff
@@ -267,9 +274,45 @@ def get_our_mint_txs(self, pair_address: Address) -> dict:
...
+                    token1Price
+                    volumeToken0
+                    volumeToken1
+                    reserve0
+                    reserve1
+                    txCount
+                    id
+                }
+            }
+        """
```
EdNoepel, line 307: Questions about the query:
1. Do we still need to query `volumeToken0` and `volumeToken1`?
2. Where can I find docs for the `pairs` query?  I looked [here](https://thegraph.com/explorer/subgraph/graphprotocol/uniswap) but didn't find these parameters.

MikeHathaway: 1. Nope, I've removed those attributes.
2. Each version of Uniswap has its own subgraph, v2 is available [here](https://thegraph.com/explorer/subgraph/uniswap/uniswap-v2). Block is available as a filter on all the top level uniswapv2 subgraph entities.

EdNoepel: I'm curious what `reserve0` and `reserve1` show.  The documentation didn't seem to offer a description of these fields.

MikeHathaway: reserve0 and reserve1 are reflection of the pair contract (https://github.com/Uniswap/uniswap-v2-subgraph/blob/master/schema.graphql#L64) (https://uniswap.org/docs/v2/smart-contracts/pair/#getreserves), and represent the the amount of each asset for a given pair locked in the contract.

pyexchange/uniswapv2_analytics.py
```diff
@@ -53,25 +52,25 @@ def from_our_trades_message(trade: dict, pair: str, base_token: Token, our_liqui
...
+        our_pool_share = our_liquidity_balance / Wad.from_number(trade['totalSupply'])
 
-        if trade['pair']['token0']['id'] == base_token.address.address:
+        if trade['token0']['id'] == base_token.address.address:
             swap_price = Wad.from_number(trade['reserve1']) / Wad.from_number(trade['reserve0'])
 
             is_sell = Wad.from_number(trade['reserve1']) < previous_base_token_reserves
 
-            amount = our_pool_share * Wad.from_number(trade['hourlyVolumeToken0'])
+            amount = our_pool_share * abs(Wad.from_number(trade['reserve1']) - previous_base_token_reserves)
```
EdNoepel, line 64: I'm curious why the `abs` was necessary; could a pool have negative tokens, and if it did, why would we pretend the value is positive?

MikeHathaway: Negative reserve's aren't possible with Uniswap or other `x * y = k` exchanges as negative number's would necessarily violate the constant product invariate. 

I was taking the `abs` in order to determine the difference in reserves between the two periods, while avoiding any issues from negative amounts in our calculations. For example, if the previous reserves were greater than i'm assuming we've essentially sold off some amount of the reserves. If less, than we've essentially bought the base currency.

EdNoepel (approved):

EdNoepel (approved):

PR #181, opened 2020-10-14
Add setuptools

MikeHathaway:


setup.py
```diff
@@ -0,0 +1,66 @@
...
+setup(
+    name='pyexchange',
+
+    # Versions should comply with PEP 440:
+    # https://www.python.org/dev/peps/pep-0440/
+    #
+    # For a discussion on single-sourcing the version across setup.py and the
+    # project code, see
+    # https://packaging.python.org/en/latest/single_source_version.html
+    version='1.1.0',  # Required
```
EdNoepel, line 31: We'll need to figure out how to increment this as time progresses.  It shouldn't be pegged to `pymaker`, but I have no problem with the initial version being 1.1.0.

setup.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # This field lists other packages that your project depends on to run.
+    # Any package you put here will be installed by pip when your project is
+    # installed, so they must be valid existing projects.
+    #
+    # For an analysis of "install_requires" vs pip's requirements files see:
+    # https://packaging.python.org/en/latest/requirements.html
+    #
+    # All packages in pyexchange.requirements.txt and pyexchange/lib/pymaker.requirements.txt
+    install_requires=[
```
EdNoepel, line 51: There's a way to reduce maintenance costs: see https://github.com/makerdao/pygasprice-client/blob/master/setup.py#L19

MikeHathaway: That is much cleaner, I'll update it accordingly.

setup.py
```diff
@@ -0,0 +1,65 @@
...
+from os import path
+
+here = path.abspath(path.dirname(__file__))
+
+# Get the long description from the README file
+with open(path.join(here, 'README.md'), encoding='utf-8') as f:
+    long_description = f.read()
+
+# Read requirements.txt
+with open(path.join(here, 'requirements.txt'), encoding='utf-8') as f:
```
MikeHathaway, line 22: this is still a WIP

pyexchange/fix.py
```diff
@@ -152,32 +185,95 @@ def write(self, message: simplefix.FixMessage):
...
+
+        cancel_message_types = [simplefix.EXECTYPE_CANCELED, simplefix.EXECTYPE_EXPIRED]
+        open_orders = []
+
+        for order in orders:
+            client_order_id = self._get_client_id(order.order_id)
+            order_status_dict = {}
+
+            while not self.order_book[client_order_id].empty():
+                exchange_order_state = self.order_book[client_order_id].get()
```
EdNoepel, line 214: Here's a list of possible states and my commentary on them:

```
EXECTYPE_NEW = b'0'		- how are these handled?
EXECTYPE_PARTIAL_FILL = b'1'	- handled
EXECTYPE_FILL = b'2'		- handled
EXECTYPE_DONE_FOR_DAY = b'3'	- needed for GTD orders
EXECTYPE_CANCELED = b'4'	- handled
EXECTYPE_REPLACE = b'5'		- probably not needed since we don't implement cxl/replace
EXECTYPE_PENDING_CANCEL = b'6'  - how are these handled?
EXECTYPE_STOPPED = b'7'		- unsure
EXECTYPE_REJECTED = b'8'	- should handle invalid orders
EXECTYPE_SUSPENDED = b'9'	- probably not needed
EXECTYPE_PENDING_NEW = b'A'	- how are these handled?
EXECTYPE_CALCULATED = b'B'	- probably not needed
EXECTYPE_EXPIRED = b'C'		- handled
EXECTYPE_RESTATED = b'D'	- probably not needed
EXECTYPE_PENDING_REPLACE = b'E' - probably not needed since we don't implement cxl/replace
```

Here's some useful documentation regarding the state machine: 
https://www.fixtrading.org/online-specification/order-state-changes/

Just want to ensure we're handling all relevant states.

pyexchange/fix.py
```diff
@@ -152,32 +185,95 @@ def write(self, message: simplefix.FixMessage):
...
-                        logging.error(f"Order placement rejected due to {message.get(58).decode('utf-8')}, tag_103 code: {message.get(103).decode('utf-8')}")
-                    return message
+            if client_order_id in self.order_book:
+                if not self.order_book[client_order_id].empty():
+                    message = self.order_book[client_order_id].get()
+                    assert isinstance(message, simplefix.FixMessage)
+
+                    # handle message rejection
+                    if message.get(simplefix.TAG_MSGTYPE) in reject_message_types:
+                        if message.get(102) is not None:
```
EdNoepel, line 264: `TAG_CXLREJREASON`

pyexchange/fix.py
```diff
@@ -152,32 +185,95 @@ def write(self, message: simplefix.FixMessage):
...
+                            logging.error(f"Order cancellation rejected due to {message.get(58).decode('utf-8')}, tag_102 code: {message.get(102).decode('utf-8')}")
+                        if message.get(103) is not None:
```
EdNoepel, line 266: `TAG_ORDERREJREASON`

EdNoepel (commented):
I'm happy with the design of the changes.  Added some comments about ensuring we're handling all potential states.  Also recommend renaming this PR, since it's not just `setuptools` anymore.

pyexchange/fix.py
MikeHathaway: Updated

pyexchange/fix.py
MikeHathaway: Updated

pyexchange/fix.py
```diff
@@ -152,32 +185,95 @@ def write(self, message: simplefix.FixMessage):
...
         pass
 
-    async def _wait_for_response(self, message_type: str) -> simplefix.FixMessage:
+    def _get_client_id(self, order_id: str) -> str:
+        """ Retrieve the client_id from an order_id structured as exchange_id|client_id """
+        assert isinstance(order_id, str)
+
+        return order_id.split('|')[1]
+
+    # TODO: return a List[{order_id: amount_remaining}]
```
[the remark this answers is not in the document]
MikeHathaway: Removed

pyexchange/fix.py
MikeHathaway: Thanks, this is very helpful. As you pointed out, pending messages aren't currently handled - this is a byproduct of the ErisX design in that they don't seem to send `PENDING_CANCEL` or `PENDING_NEW` messages. Additionally, some orders can be rejected with `35=8` and non-standard `103=100` or `103=23` codes. I just added additional checks for these as well.

EdNoepel (approved):
LGTM

pyexchange/fix.py
MikeHathaway: 'EXECTYPE_NEW' is handled by the catchall message return on L269 of `_wait_for_order_processing_response()`. `EXECTYPE_DONE_FOR_DAY` is not currently supported by ErisX. In the event that an Order is cancelled by ErisX, it will be recorded as a standard Order Cancellation Exectuion Report, which sync_orders should catch and remove from our client's order book.

PR #183, opened 2020-10-23
Fix BinanceUS order cancelation error

Exef:


ith-harvey (approved):
* fixed cancellations. Looks good and has been tested.

PR #184, opened 2020-10-28
Add Gemini Exchange integration

Exef:
Documentation: https://docs.gemini.com/rest-api


pyexchange/gemini.py
```diff
@@ -0,0 +1,215 @@
...
+    trades = self._http_authenticated("POST", "/v1/mytrades", {"symbol": self._fix_pair(pair)})
+    return list(map(lambda trade: GeminiTrade.from_my_trade(pair, trade), trades))
+
+  def get_all_trades(self, pair: str, page_number: int = 1) -> List[GeminiTrade]:
+    assert(isinstance(pair, str))
+    assert(isinstance(page_number, int))
+    
+    all_trades = self._http_unauthenticated("GET", f"/v1/trades/{self._fix_pair(pair)}")
+
+    return list(map(lambda trade: GeminiTrade.from_trade(pair, trade), all_trades))
```
MikeHathaway, line 151: Does all trades endpoint paginate?

pyexchange/gemini.py
```diff
@@ -0,0 +1,215 @@
...
+      "side": "sell" if is_sell else "buy",
+      "type": "exchange limit"
+    }
+    placed_order = self._http_authenticated("POST", "/v1/order/new", params)
+
+    return placed_order["order_id"]
+
+  def cancel_order(self, order_id: str, pair: str) -> bool:
+    assert(isinstance(order_id, str))
+    self.logger.info(f"Cancelling order #{order_id} on pair {pair}...")
```
MikeHathaway, line 132: Is `pair` needed here? There's already order cancellation logging in `market-maker-keeper`, and this will lead to additional code there to override `cancel_orders_with` interface.

pyexchange/gemini.py
```diff
@@ -0,0 +1,215 @@
...
+    assert(isinstance(api_key, str))
+    assert(isinstance(api_secret, str))
+    assert(isinstance(timeout, float))
+
+    self.api_server = api_server
+    self.api_key = api_key
+    self.api_secret = api_secret
+    self.timeout = timeout
+
+  def get_balances(self):
```
MikeHathaway, line 86: Is there a reason this returns a dictionary? `Market-Maker-Keeper` expects a `List` I believe. Also, adding the return type would be helpful for debugging.

pyexchange/gemini.py
Exef: No, it returns up to 500 trades up to 7 days ago. `since` timestamp can be also defined

https://docs.gemini.com/rest-api/#trade-history

pyexchange/gemini.py
Exef: As far as I know, it's unspecified what this method should return. Usually, it returns an unchanged response from an API, and let's keeper handle it.

pyexchange/gemini.py
Exef: Removed.

pyexchange/gemini.py
MikeHathaway: I've actually had to convert it from `dict` into `List` before in DyDx pyexchange, as a `dict` is never expected. You can of course handle it in the consumer, but you will then need to make the identical conversion to a `list` in both `inventory-service`, and `market-maker-keeper`.

MikeHathaway (approved):
PR generally looks good, `get_balances` would be more efficient as a list, but that isn't a big deal.

PR #185, opened 2020-10-29
Fix ErisX order cancellation monopolization

MikeHathaway:
mismatch between `TAG_ORIGCLORDID` and `TAG_CLORDID` caused new entries to be added in the orderbook, as opposed to the cancellation of existing orders. This PR fixes the issue, and enables orders to be cancelled without restarting the keeper.


pyexchange/fix.py
```diff
@@ -137,7 +137,7 @@ def _handle_application_message(self, message: simplefix.FixMessage):
 
         # handle order processing messages
         if message.get(simplefix.TAG_MSGTYPE) == simplefix.MSGTYPE_EXECUTION_REPORT:
-            client_order_id = f"{message.get(simplefix.TAG_CLORDID).decode('utf-8')}"
+            client_order_id = f"{message.get(simplefix.TAG_ORIGCLORDID).decode('utf-8')}"
```
EdNoepel, line 140: You should only expect tag 41 when an order is cancelled or cxl/replaced (the latter we don't support).  So this doesn't look right.  Are you seeing message 8s arrive without tag 11?

EdNoepel (changes requested):
Lets discuss tomorrow.

pyexchange/fix.py
MikeHathaway: Yeah this is another one of the idiosyncrasies of ErisX. Even newly placed orders respond with tag 41. Only, in that case, tag 41 == tag 11.

The problem was that order cancellation tag 11 differed from the existing client_id, so the new messages couldn't be appended to the existing queue.

EdNoepel: So for a cancellation message, what do they send in tag 11?  I'd be happy to review logs with you.

I suspect we should consume tag 11 or 41 based on `TAG_EXECTYPE`.

EdNoepel: Per our discussion, I recommend consuming tag 41 if 150 is 4,5,6 or E, and tag 11 otherwise.  

Structurally, I recommend the _outer_ conditional handle each supported tag 150 use case, with each `elif` pulling `clorderid` from the appropriate field before changing the state of the local order book.

EdNoepel (approved):

PR #188, opened 2020-11-11
Add staking rewards

MikeHathaway:


pyexchange/uniswap_staking_rewards.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(self, web3: Web3, keeper_address: Address, contract_address: Address, contract_name: str):
+        assert (isinstance(web3, Web3))
+        assert (isinstance(keeper_address, Address))
+        assert (isinstance(contract_address, Address))
+        assert (isinstance(contract_name, str))
+
+        self.web3 = web3
+
+        self.keeper_address = keeper_address
+        self.contract_address = Address(contract_address)
```
EdNoepel, line 52: Why does the `Address` ctor need to be used here?

pyexchange/uniswap_staking_rewards.py
```diff
@@ -0,0 +1,86 @@
...
+        super().__init__(web3, self.keeper_address, self.contract, self.contract_abi, self.contract_address, self.contract_name)
+
+    def stake_liquidity(self, amount: Wad) -> Transact:
+        assert (isinstance(amount, Wad))
+
+        stake_liquidity_args = [
+            amount.value
+        ]
+        print(self.contract_address)
+        print((isinstance(Address(self.contract_address), Address)))
```
EdNoepel, line 66: These `print` statements could be combined into a single log message if needed.

pyexchange/uniswap_staking_rewards.py
```diff
@@ -0,0 +1,86 @@
...
+        assert (isinstance(contract_address, Address))
+        assert (isinstance(contract_name, str))
+
+        self.web3 = web3
+
+        self.keeper_address = keeper_address
+        self.contract_address = Address(contract_address)
+        self.contract = self._get_contract(self.web3, self.staking_rewards_abi, self.contract_address)
+        self.contract_abi = self.staking_rewards_abi
+        self.contract_name = contract_name
```
EdNoepel, line 55: It's awkward assigning these five vars here, and then passing them as args to `super().__init__` which reassigns them.  Recommend either:
- making them local vars and allowing the superclass to perform the assignment
- initialize them directly in the `super().__init__` call

pyexchange/uniswap_staking_rewards.py
```diff
@@ -0,0 +1,86 @@
...
+    def stake_liquidity(self, amount: Wad) -> Transact:
+        assert (isinstance(amount, Wad))
+
+        stake_liquidity_args = [
+            amount.value
+        ]
+        print(self.contract_address)
+        print((isinstance(Address(self.contract_address), Address)))
+        # TODO: what is goig on with contract address no longer being an Address type?
+        return Transact(self, self.web3, self.contract_abi, Address(self.contract_address), self.contract,
```
EdNoepel, line 68: That explains my previous question about the weird cast.  Was it reverting to a string?  I don't see anything which would cause such behavior.

pyexchange/uniswap_staking_rewards.py
```diff
@@ -0,0 +1,86 @@
...
+        withdraw_liquidity_args = [
+            amount.value
+        ]
+
+        return Transact(self, self.web3, self.contract_abi, self.contract_address, self.contract,
+                'withdraw', withdraw_liquidity_args)
+
+    def withdraw_all_liquidity(self) -> Transact:
+        return Transact(self, self.web3, self.contract_abi, self.contract_address, self.contract,
+                        'exit', [])
```
EdNoepel, line 86: nit: newline at EOF

EdNoepel (commented):
Comments within.  Appreciate the unit tests.

pyexchange/uniswap_staking_rewards.py
MikeHathaway: Removed both

pyexchange/uniswap_staking_rewards.py
MikeHathaway: Thanks for the suggestion. Opted for the first recommendation, which appears much cleaner.

pyexchange/uniswap_staking_rewards.py
MikeHathaway: It didn't was left over from the mangled `__init__`

pyexchange/uniswap_staking_rewards.py
MikeHathaway: It seems to have been an issue with the double assignment to `self.contract_address`. Fixed by setting in the superclass.

pyexchange/uniswap_staking_rewards.py
MikeHathaway: Added

EdNoepel (approved):

PR #189, opened 2020-11-12
this small change ensures, our bids or offers can't cross leverj futu…

mitakash:
…res index value by 85 bps, eliminates a potential arb. This has been tested on leverj futures kovan site kovan.leverj.io for both BTC and ETH futures


MikeHathaway (commented):
should this be passed through as an argument, or is this going to stay constant?

MikeHathaway:
> should this be passed through as an argument, or is this going to stay constant?

Answered in standup

MikeHathaway (approved):
LGTM

PR #190, opened 2020-11-15
Add method for resetting ErisX FIX trading passwords

MikeHathaway:


pyexchange/erisx.py
```diff
@@ -95,17 +95,18 @@ def __init__(self, fix_trading_endpoint: str, fix_trading_user: str,
         if fix_trading_endpoint is not None and fix_trading_user is not None:
             self.fix_trading = ErisxFix(fix_trading_endpoint, fix_trading_user, fix_trading_user, password, certs)
             self.fix_trading.logon()
-            self.fix_trading_user = fix_marketdata_user
+            self.fix_trading_user = fix_trading_user
 
         if fix_marketdata_endpoint is not None and fix_marketdata_user is not None:
-            self.fix_marketdata = ErisxFix(fix_marketdata_endpoint, fix_marketdata_user, fix_trading_user, password,
+            self.fix_marketdata = ErisxFix(fix_marketdata_endpoint, fix_marketdata_user, fix_marketdata_user, password,
```
EdNoepel, line 101: Why does the `ErisxFix` constructor require two usernames?

MikeHathaway: One is for `senderCompId` (for logging I believe) one is for the actual connection username. That being said, they seem to always be the same so we could simplify it by making `senderCompId == username`

EdNoepel (approved):
LGTM

PR #192, opened 2020-11-18
Add choose nonce to pyex

MikeHathaway:


EdNoepel (approved):
Of course, this will only synchronize across threads, not across processes.

pyexchange/api.py
```diff
@@ -10,6 +12,10 @@ class PyexAPI:
     Define a common abstract API for exchanges
     """
 
+    last_nonce = 0
+    last_nonce_lock = threading.Lock()
```
ith-harvey, line 16: I believe this needs to be `self.last_nonce_lock`

pyexchange/api.py
```diff
@@ -10,6 +12,10 @@ class PyexAPI:
     Define a common abstract API for exchanges
     """
 
+    last_nonce = 0
```
ith-harvey, line 15: I believe this needs to be `self.last_nonce`

ith-harvey (commented):
This is great Mike, think you need to change some var names. Could you also please incorporate this into Gemini? 

Or upon pushing this PR we could get @Exef to make the change to Gemini since he already has the test env setup.

pyexchange/api.py
EdNoepel: Good point; that's the difference between class and instance variables.  If he made these instance variables, there would be a different nonce for each instance of the class, which might not be what is desired.

As implemented they are class variables, which means the lock would be shared across instances of the class.  If they were implemented in a subclass, that would make sense, because multiple instances for that exchange should share the same lock, not have their own.  But here the class variable is in the baseclass.  If the user has two `PyexAPI` instances which are for different exchanges (read: different subclases), they shouldn't share the same nonce.

ith-harvey (approved):
LGTM thanks @EdNoepel for the great explanation

PR #195, opened 2020-12-05
handle erisx orders that have been filled before they can be cancelled

MikeHathaway:


EdNoepel:
Could you provide an explanation of the change?  The code this PR removes seems to remove filled orders from the local order book, which would presumably help prevent attempts to cancel them.

MikeHathaway:
> Could you provide an explanation of the change? The code this PR removes seems to remove filled orders from the local order book, which would presumably help prevent attempts to cancel them.

Sure, so one issue was that the method used to assign incoming messages to a given orders queue was incomplete. Rejected Order Cancel Requests are returned with `35=9`, not `35=8` so they weren't captured during `sync_orders`. 

The second issue, and the cause of the restarts, was due to timing. The price would shift and exceed band targets, prompting the client side order book to send an order cancellation request. Before this cancellation could be processed, the order would be filled. Given the way we make order cancellations synchronous, this order fill would instead be handled by the waiting `cancel_order` method. It was not previously structured to handle filled orders in the circumstance, so it would tell the client (mm-keeper) order book that the cancellation failed, prompting it to retry. This retry would then hang forever, due to the first issue where the rejection message wasn't properly queued for that order id.

EdNoepel (approved):
That makes sense; thank you.

PR #197, opened 2020-12-15
increase timeout to see if it fixes read timeout fetching order book

EdNoepel:
One-liner to be less aggressive with the timeout.


MikeHathaway (approved):
LGTM

PR #201, opened 2021-01-04
Liquid nonce solution

ith-harvey:
This solution fixes the nonce issue we saw at liquid by locking the thread the entire time (from when nonce is generated -> when the request is sent)


EdNoepel (approved):
I'm Ed Noepel and I approve this message.

PR #203, opened 2021-02-16
handle unsolicited order cancellations

MikeHathaway:


EdNoepel (approved):
Confirmed 5001 isn't a standard supported constant in simplefix.  Looks good.

PR #204, opened 2021-02-18
Lock order deletion from local order book

MikeHathaway:
Logs were showing key errors from attempting to delete entries that were no longer present. This should resolve that issue by ensuring that orders are removed from the orderbook on receipt of cancellation or fill.

### DIFF
--- a/pyexchange/fix.py
+++ b/pyexchange/fix.py
@@ -281,6 +281,9 @@ class FixEngine:
                         # Remove orders that have been cancelled from the order book
                         if message.get(simplefix.TAG_EXECTYPE) == simplefix.EXECTYPE_CANCELED:
                             del self.order_book[client_order_id]
+                        # Remove orders that recieve an UnsolicitedCancel response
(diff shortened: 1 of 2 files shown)


pyexchange/fix.py
```diff
@@ -194,6 +194,17 @@ def write(self, message: simplefix.FixMessage):
         self.write_queue.put(message)
         pass
 
+    async def _delete_order(self, client_order_id: str):
+        assert (isinstance(client_order_id, str))
+
+        await self.lock.acquire()
+        try:
+            del self.order_book[client_order_id]
+        except Exception as e:
```
EdNoepel, line 203: I'd usually gripe about the catch-all, but in this case it seems appropriate.

EdNoepel (approved):
The threadsync logic protects against multiple concurrent calls of `_sync_orders`.  The private deletion method is an elegant way to safely delete without having a `KeyError` interrupt the application.

PR #205, opened 2021-02-26
add lock to fix.py when processing incoming order messages

MikeHathaway:


MikeHathaway:
Attempting to resolve an issue wherein an order is placed with a successful response, immediately followed by a second response message that the successfully placed order would violate take only requirements. This edge case should be currently handled by checking `messge.get(b'5001') is not None, since all such INVALID ALO or cancellation rejection messages contain this field.

Since the shutdowns are still occuring, even with this check on both `sync_orders` and `wait_for_order_processing_response`, this is making me think there is an asynchronicity issue from these reject messages not being queued prior to the unnecessary attempted cancellation.

pyexchange/fix.py
```diff
@@ -112,7 +112,9 @@ async def _read_message(self):
 
             # Handle session messages, queue application messages.
             if not self._handle_session_message(message):
+                await self.lock.acquire()
                 self._handle_application_message(message)
+                self.lock.release()
```
EdNoepel, line 117: Recommend calling `_handle_application_message` in a `try` and releasing the lock in a `finally`.  A catch isn't needed.

MikeHathaway: Thanks for the suggestion, updated accordingly

EdNoepel (approved):

PR #206, opened 2021-03-02
Altered place_order for DYDX token bot

ith-harvey:
* altered place order method:
  * added `fee_limit` and `fill_or_kill` param
* added separate `_set_allowance` method
* added a `get_all_orders` method


EdNoepel (approved):
LGTM

